PR #20, opened 2021-10-12
Create carma_to_carla_ackermann_cmd

chengyuan0124:
Upload carma_to_carla_ackermann_cmd script

# PR Details
## Types of changes

## Checklist:

### DIFF
--- a/carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carla_to_autoware_vehicle_status
+++ b/carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carla_to_autoware_vehicle_status
@@ -60,9 +60,8 @@ def vehicle_status_callback(vehicle_status_msg):
 
     # get max steering angle (use smallest non-zero value of all wheels)
     for wheel in vehicle_info.wheels:
-        if wheel.max_steer_angle:
-            if wheel.max_steer_angle and wheel.max_steer_angle < max_steering_angle:
-                max_steering_angle = wheel.max_steer_angle
+        if wheel.max_steer_angle and wheel.max_steer_angle < max_steering_angle:
+            max_steering_angle = wheel.max_steer_angle
 
     status.angle = vehicle_status_msg.control.steer * math.degrees(max_steering_angle)
     status.speed = vehicle_status_msg.velocity
--- a/carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carla_to_autoware_vehicle_status
+++ b/carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carla_to_autoware_vehicle_status
@@ -34,6 +34,7 @@ from autoware_msgs.msg import VehicleStatus, Gear
 pub = rospy.Publisher('/hardware_interface/vehicle_status', VehicleStatus, queue_size=1)
 vehicle_info = None
 
+max_steering_angle = rospy.get_param('/max_steering_angle', 70)
 
 def vehicle_info_callback(vehicle_info_msg):
     """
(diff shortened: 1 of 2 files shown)


carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_to_carla_ackermann_cmd
```diff
@@ -0,0 +1,110 @@
...
+        ackermann_cmd_pub.publish(init_cmd)
+
+def commanded_jerk_callback(lateral_jerk_msg):
+    """
+    callback for commanded jerk from CARMA
+    lateral_jerk_msg type:
+        std_msgs::Float32
+    """
+
+    global jerk
```
kjrush, line 69: This jerk variable is created (as a reference to a global) but never used here. Either we should delete this line, or perhaps the line after needs a change to reference this variable?

kjrush: Actually, it looks like there's a mismatch, insofaras the global variable exists, but it's called `lateral_jerk` not `jerk`. So I think this line needs an update to match the global variable name.

kjrush (changes requested):

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_to_carla_ackermann_cmd
chengyuan0124: Didn't notice about that, I've updated to match the global variable name

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carla_to_autoware_vehicle_status
```diff
@@ -0,0 +1,97 @@
...
+    vehicle_status_msg type:
+        carla_msgs::CarlaEgoVehicleStatus
+    """
+    if vehicle_info is None:
+        return
+    status = VehicleStatus()
+    status.header = vehicle_status_msg.header
+
+    # calculate max steering angle
+    max_steering_angle = math.radians(70)
```
kjrush, line 59: This might be best left as a config value, different vehicles may have different max steering angles.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carla_to_autoware_vehicle_status
```diff
@@ -0,0 +1,97 @@
...
+
+    # get max steering angle (use smallest non-zero value of all wheels)
+    for wheel in vehicle_info.wheels:
+        if wheel.max_steer_angle:
+            if wheel.max_steer_angle and wheel.max_steer_angle < max_steering_angle:
+                max_steering_angle = wheel.max_steer_angle
+
+    status.angle = vehicle_status_msg.control.steer * math.degrees(max_steering_angle)
```
kjrush, line 67: Is `vehicle_status_msg.control.steer` a value in range [-1.0, 1.0]? Do we need to explicitly clamp it into that range or are we safe to assume it'd never be outside of those values?

Also, just for clarity sake, this is wheel angel not steering wheel angle correct?

chengyuan0124: The `vehicle_status_msg.control.steer` is defined in range [-1.0, 1.0] according to the CARLA official [website](https://carla.readthedocs.io/en/0.9.8/ros_msgs/#carlaegovehiclecontrolmsg), so it will not be outside of range.

The script `status.angle = vehicle_status_msg.control.steer * math.degrees(max_steering_angle)` is for calculating wheel angle and set into `VehicleStatus` for CARMA.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carla_to_autoware_vehicle_status
chengyuan0124: I've updated the script for leaving max_steering_angle parameter as a config value

kjrush (approved):

PR #49, opened 2022-02-18
Feature/carla carma state integration

lewisid:
# PR Details
## Types of changes

## Checklist:


carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_to_carla_route
```diff
@@ -0,0 +1,130 @@
...
+class ErrorStatus:
+    def __init__(self, value, text):
+        self.value = value
+        self.text = text
+
+
+def get_available_routes():
+    service = rospy.ServiceProxy('/guidance/get_available_routes', GetAvailableRoutes)
+    #Make sure that the vehicle is localized
+    if localized_sub:
```
chengyuan0124, line 45: There might not need to check the localized before getting available routes.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_to_carla_route
```diff
@@ -0,0 +1,130 @@
...
+def route_event_callback(event):
+    route_event = RouteEvent()
+    route_event.event = event
+    if route_event.event == 6:
+        print("ROUTE_GEN_FAILED")
+    
+
+def initialize():
+    available_routes = get_available_routes()
+    route = available_routes.availableRoutes[0]
```
chengyuan0124, line 119: It might need a parameter/config file to indicate which route id is going to use instead of always selected first one

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_plugins
```diff
@@ -0,0 +1,100 @@
...
+    for i in required_plugins:
+        if i == name:
+            return i
+        else:
+            return FALSE
+
+def initialize():
+    plugin_sub = rospy.Subscriber('plugin/plugin_discovery',Plugin(), update_plugin_status)
+    #registered_plugins_serv = rospy.ServiceProxy('carla/get_registered_plugins',get_registered_plugins)
+    registered_plugins_serv = rospy.Service('carla/get_registered_plugins',GetRegisteredPlugins,get_registered_plugins)
```
kjrush, line 94: This is advertising a new service right? I believe the intent would be to call the CARMA Platform's get_registered_plugins service instead, so I think a `ServiceProxy` object is more appropriate.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_plugins
```diff
@@ -0,0 +1,100 @@
...
+import rospy
+from cav_msgs.msg import Plugin, PluginList, PluginStatus
+from cav_srvs.srv import GetRegisteredPlugins
+
+
+plugin_pub = rospy.Publisher('/guidance/plugin_discovery', Plugin())
+required_plugins = rospy.get_param('required_plugins')
+plugin_list = PluginList()
+
+def get_registered_plugins(resp):
```
kjrush, line 38: It's not clear to me what this function is doing, it looks like it's expecting to take a service call response as input (judging from the name `resp`) but the `get_registered_plugins` service response from the Guidance Health Monitor node already contains all the plugin information for all registered plugins.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_plugins
```diff
@@ -0,0 +1,100 @@
...
+    resp.plugins.append(plugin)
+        #Publish the now active plugin message to /guidance/plugin_discovery
+        #plugin_pub.publish(plugin)
+        
+
+#Check if the plugin is included in the list of required plugins
+def check_required_plugins(name):
+    for i in required_plugins:
+        if i == name:
+            return TRUE
```
kjrush, line 62: I think all the boolean values in this PR should be replaced with the native Python `True`/`False` literals instead of the ones from the Pickle library.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_plugins
```diff
@@ -0,0 +1,100 @@
...
+        else:
+            return FALSE
+
+def update_plugin_status(plugin_msg):
```
kjrush, line 66: Is there a reason to maintain a list of plugin states here? It looks to me like this function is processing a message from the `plugin_registration` topic where the plugins are publishing their status and updating a global list correct? It's not clear what that list is needed for to me, though.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_plugins
```diff
@@ -0,0 +1,100 @@
...
+    if requested_name != FALSE:
+        plugin.activated = plugin_msg.activated
+    elif check_required_plugins(plugin_msg.name) == TRUE:
+        plugin.activated = TRUE
+
+    plugin_list.plugins.append(plugin)
+
+
+def get_plugin_name(name):
+    for i in required_plugins:
```
kjrush, line 85: I think what you're looking for in this function can be accomplished with `return name in required_plugins` or something along those lines, Python's `in` keyword can do membership tests and return True or False. Currently this is returning the input name, so maybe `return name if name in required_plugins else None` if that's more desirable from the caller's perspective.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_plugins
```diff
@@ -0,0 +1,100 @@
...
+        if i == name:
+            return i
+        else:
+            return FALSE
```
kjrush, line 89: Returning different types depending on the logic is probably not best practice, for such cases like this it's probably best to return `None` (similar to nullptr in C++) to represent a fail-to-find.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_plugins
```diff
@@ -0,0 +1,100 @@
...
+def update_plugin_status(plugin_msg):
+    print("Received status from: " ++ plugin_msg.name)
+    requested_name = get_plugin_name(plugin_msg.name)
+    plugin = Plugin()
+    plugin.available = plugin_msg.available
+    plugin.name = plugin_msg.name
+    plugin.capability = plugin_msg.capability
+    plugin.type = plugin_msg.type
+
+    #If the requested plugin exists, don't change the activation status
```
kjrush, line 75: Is this where we'd be wanting to actually activate the plugin? To do so you'd need to make a call to the service defined here: https://github.com/usdot-fhwa-stol/carma-platform/blob/444af8eb57c57b7b91237e81b554e284407d4f13/health_monitor/src/health_monitor.cpp#L39 so that the health_monitor can actually activate/deactivate the plugin. The Health Monitor and the plugin itself are the actual owners of that data, the get_registered_plugins service and plugin_discovery topic are just outputs that reflect the current state. Those topic/service don't actually control the underlying data.

kjrush (changes requested):

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
```diff
@@ -0,0 +1,146 @@
...
+    if available_routes == None:
+        rospy.spin()
+    else:
+        """Check if the selected route is one of the available routes. If it isn't,then use the first of the available 
+        routes received from the service call"""
+        if check_selected_route(available_routes) == True:
+            route = selected_route
+        else:
+            route = available_routes.availableRoutes[0]
+        set_route(route.route_id)
```
chengyuan0124, line 136: The set_route function should be called after CARMA received the vehicle position data from CARLA. You could subscribe the ROS topic of `/localization/current_pose` to see whether the position data is available or not

carla-carma-integration/carla_carma_bridge/CMakeLists.txt
```diff
@@ -8,7 +8,10 @@ project(carla_carma_bridge)
 find_package(catkin REQUIRED COMPONENTS
   autoware_msgs
   rospy
+  cav_srvs
   cav_msgs
+  health_monitor
```
kjrush, line 13: Is this dependency on health_monitor necessary? I don't see it being used anywhere

carla-carma-integration/carla_carma_bridge/package.xml
```diff
@@ -16,6 +16,9 @@
   <exec_depend>autoware_msgs</exec_depend>
   <exec_depend>carla_msgs</exec_depend>
   <exec_depend>cav_msgs</exec_depend>
+  <exec_depend>cav_srvs</exec_depend>
+  <exec_depend>health_monitor</exec_depend>
```
kjrush, line 20: Same here with the dependency

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_plugins
```diff
@@ -0,0 +1,82 @@
...
+#
+# This work is licensed under the terms of the MIT license.
+# For a copy, see <https://opensource.org/licenses/MIT>.
+#
+
+import rospy
+from cav_msgs.msg import Plugin, PluginList, PluginStatus
+from cav_srvs.srv import GetRegisteredPlugins
+
+plugin_pub = rospy.Publisher('/guidance/plugin_discovery', Plugin())
```
kjrush, line 32: I think this publish can be removed per our most recent discussions right? The subscriber below is handling what's needed I believe.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_plugins
```diff
@@ -0,0 +1,82 @@
...
+selected_plugins = rospy.get_param('selected_plugins')
+
+#List to contain the registered plugins from the service call in initialize()
+plugin_list = PluginList()
+
+
+#Check if the plugin is included in the list of required plugins
+def check_selected_plugins(name):
```
kjrush, line 40: I think this function is equivalent to `name in selected_plugins` if you want to simplify things a bit.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_plugins
```diff
@@ -0,0 +1,82 @@
...
+    for i in selected_plugins:
+        if i == name:
+            return True
+        else:
+            return False
+
+def plugin_status_cb(plugin_msg):
+    print("Received status from: " ++ plugin_msg.name)
+    requested = get_plugin(plugin_msg)
+    plugin = Plugin()
```
kjrush, line 50: Is the intent to do something with this plugin object? It looks like it gets created here in the callback, some data gets copied into it from the callback data, and then it falls out of scope and is deleted. If the intent is to save the data for later this probably needs to get added to a list somewhere with a longer lifetime (maybe the global plugin_list?). Or if it's just for temporary use, I think we can do away with it entirely and just use the data directly from the callback.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_plugins
```diff
@@ -0,0 +1,82 @@
...
+    #If the requested plugin is registered, don't change the activation status
+    if requested != None:
+        plugin.activated = plugin_msg.activated
+    elif check_selected_plugins(plugin_msg.name) == True:
+        plugin.activated = True 
+    activation_srv.call(plugin.name, plugin.activated)
+
+
+#Search through list of registered plugins. If it exists, return the requested plugin
+def get_plugin(plugin):
```
kjrush, line 66: Can you explain a little bit of the intent here? We pass in a plugin object (containing all the plugin data fields) and if that exact object exists already in the plugin list, we get that object back, otherwise we get None? Is the intent for this to check if we've seen data about this plugin already?

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_plugins
```diff
@@ -0,0 +1,82 @@
...
+    plugin.capability = plugin_msg.capability
+    plugin.type = plugin_msg.type
+    activation_srv = rospy.ServiceProxy("/activate_plugin")
+
+    #If the requested plugin is registered, don't change the activation status
+    if requested != None:
+        plugin.activated = plugin_msg.activated
+    elif check_selected_plugins(plugin_msg.name) == True:
+        plugin.activated = True 
+    activation_srv.call(plugin.name, plugin.activated)
```
kjrush, line 62: I think we probably want to just call this service once per plugin. We'll be getting these plugin status updates at ~10hz I think, and while I don't think it'll necessarily cause issues to call the service repeatedly with the same data, it's not how the service is intended to be used. It's probably best to call it maybe only the first time we see a given plugin report that it's registered, or maybe even just directly after our initial call to the get_registered_plugins service.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
```diff
@@ -0,0 +1,146 @@
...
+# the License.
+"""
+Call Services from CARMA: 
+    Service: /guidance/get_available_routes
+             /guidance/set_active_route
+
+Publish to CARMA :cav_msgs::Route
+    
+"""
+from asyncio.windows_events import NULL
```
kjrush, line 23: I think the types on line 23-25 should be looked at, this are pretty abnormal python types so I'm not sure where they're coming into play here. Windows API stuff and the XML packages definitely aren't needed for this, and I'm not sure what the "typing" package is.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
```diff
@@ -0,0 +1,146 @@
...
+    except:
+        print("Invalid: /guidance/get_available_routes failed.")
+    else:
+        print("Available Routes acquired")
+        return available_routes.availableRoutes
+
+
+
+def set_route(id):
+    state = GuidanceState()
```
kjrush, line 61: Doesn't look like this state variable is used in this function.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
```diff
@@ -0,0 +1,146 @@
...
+    service = rospy.ServiceProxy('/set_active_route', SetActiveRoute)
+    
+    NO_ERROR = ErrorStatus(0, "NO_ERROR")
```
kjrush, line 64: I'd recommend moving these constants out of this function to be global constants instead.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
```diff
@@ -0,0 +1,146 @@
...
+    ALREADY_FOLLOWING_ROUTE = ErrorStatus(2, "ALREADY_FOLLOWING_ROUTE")
+    ROUTE_FILE_ERROR = ErrorStatus(3, "ROUTE_FILE_ERROR")
+    ROUTING_FAILURE = ErrorStatus(4, "ROUTING_FAILURE")
+    TRANSFORM_ERROR = ErrorStatus(5, "TRANSFORM_ERROR")
+
+    try:
+        #Service Call Request
+        resp = service(0, id)
+        if resp.errorStatus != NO_ERROR.value:
+            errorDescription = ""
```
kjrush, line 75: Is this errorDescription used anywhere? Not sure what all this error handling is intended for, it can probably be simplified to a check for the "NO_ERROR" value and a generic exception or error notice on all other cases. I don't think there's a meaningful way for this node to handle any of these other errors outside of maybe waiting a brief delay and retrying right?

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
```diff
@@ -0,0 +1,146 @@
...
+                #Change Guidance from READY to ACTIVE
+                #set_guidance_active()
+                   
+
+
+    except:
+            print("Fail")
+    
+    #RouteEvent
+    route_event_sub = rospy.Subscriber('/guidance/route_event', RouteEvent,route_event_callback)
```
kjrush, line 106: I'd move this subscriber setup to something like the initialize method, just so that it's setup somewhere that's more likely to only be called once.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
```diff
@@ -0,0 +1,146 @@
...
+
+
+def route_event_callback(event):
+    route_event = RouteEvent()
+    route_event.event = event
+    if route_event.event == 6:
+        print("ROUTE_GEN_FAILED")
+        
+def check_selected_route(available_routes):
+    if selected_route in available_routes:
```
kjrush, line 116:
```
  if A:
    return True
  else
    return False
```
can always be simplified to `return A` since A already has to evaluate to either a truthy value or falsey value for the if statement to work.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
```diff
@@ -0,0 +1,146 @@
...
+        return True
+    else:
+        return False
+
+"""TODO: Listen to a topic indicating that CARLA Sim has started before seting guidance to active
```
kjrush, line 121: Any progress on this todo? I think maybe the `/clock` topic might be a decent place to start but I'd double check that with @chengyuan0124

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
```diff
@@ -0,0 +1,146 @@
...
+def set_guidance_active():
+    service = rospy.ServiceProxy('/set_active_guidance')    """
+
+def initialize():
+    available_routes = get_available_routes()
+    if available_routes == None:
+        rospy.spin()
```
kjrush, line 128: This is going to infinite loop the node right? If there's no routes (for example, maybe they haven't finished loading yet) then we'll hit this line and never get to check for any routes again. I'd recommend re-working this logic.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
```diff
@@ -0,0 +1,146 @@
...
+    else:
+        """Check if the selected route is one of the available routes. If it isn't,then use the first of the available 
+        routes received from the service call"""
+        if check_selected_route(available_routes) == True:
+            route = selected_route
+        else:
+            route = available_routes.availableRoutes[0]
```
kjrush, line 135: I think for this we'd probably be better of throwing an exception or logging an error (or publishing system FATAL?) since the user's intent (run this route) can't be executed.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
kjrush: Would the current pose data then be a good topic to look at to see if the whole simulation is started up? Or is there a better cue like the clock topic? Looking for a good topic to check to see when the first simulated timestep has started so we can trigger guidance engage.

kjrush (changes requested):

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_guidance
```diff
@@ -32,14 +32,18 @@ def check_plugin_status(plugin_msg):
...
-    plugin_sub = rospy.Subscriber("/plugin_discovery", Plugin(), check_plugin_status)
-    if route_selected == True:
-        if plugins_activated == True:
-            service.call()
-    else:
-        rospy.spin()
+    #While-loop monitors the route and plugin status and sets guidance to active once both are confirmed true
+    while rospy.is_shutdown() != True:
+        route_sub = rospy.Subscriber("/route_event",RouteEvent(), check_route_status)
+        plugin_sub = rospy.Subscriber("/plugin_discovery", Plugin(), check_plugin_status)
```
chengyuan0124, line 40: The subscribers should be initialized out side of the loop

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_guidance
```diff
@@ -32,14 +32,18 @@ def check_plugin_status(plugin_msg):
...
+        if route_selected == True:
```
chengyuan0124, line 41: It could be simplified by adding `and` here. For example, `if route_selected and plugins_activated`

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_guidance
```diff
@@ -0,0 +1,45 @@
...
+from cav_msgs.msg import RouteEvent
+from cav_msgs.msg import Plugin
+
+route_selected = False
+plugins_activated = False
+
+def check_route_status(route_event_msg):
+    if route_event_msg.event == 1:
+        print("Route Selected")
+        route_selected = True
```
chengyuan0124, line 26: The `route_selected` should be set as global parameter before assign value since the `check_route_status` is a callback function.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_guidance
```diff
@@ -0,0 +1,50 @@
...
+    #While-loop monitors the route and plugin status and sets guidance to active once both are confirmed true
+    while rospy.is_shutdown() != True:
+        route_sub = rospy.Subscriber("/route_event",RouteEvent(), check_route_status)
+        plugin_sub = rospy.Subscriber("/plugin_discovery", Plugin(), check_plugin_status)
+        if route_selected == True:
+            if plugins_activated == True:
+                service.call(True) #Sets guidance status to active
+                break
+    
+    rospy.spin()
```
chengyuan0124, line 46: I think the rospy.spin() could be removed since the loop takes care of its work.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
```diff
@@ -0,0 +1,147 @@
...
+        return selected_route
+
+def initialize():
+    rospy.init_node("carma_carla_route")
+    route_event_sub = rospy.Subscriber('/guidance/route_event', RouteEvent,route_event_callback)
+    available_routes = get_available_routes()
+    
+    #Ensure that localization is complete before checking route information
+    localized_sub = rospy.Subscriber('/localization/current_pose', PoseStamped)
+    if localized_sub:
```
chengyuan0124, line 122: This condition only checks the `localized_sub` is None or not but what we actually need is to know, is there any ROS node publishing the data to the ROS topic of `/localization/current_pose`.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
```diff
@@ -122,23 +120,25 @@ def initialize():
...
-        if available_routes == None:
-            print("No routes available.")
-        else:
-            """Check if the selected route is one of the available routes. If it isn't,then print error statement"""
-            if check_selected_route(available_routes) == True:
-                route = selected_route
-                set_route(route.route_id)
-
+    while rospy.is_shutdown() != True:  
+        if local_pos.pose.position.x != 0 and local_pos.pose.position.y != 0:
```
chengyuan0124, line 124: It is better if you could initialize the value for `local_pos.pose.position.x` and `local_pos.pose.position.y` as a infinite such as `float("inf")` because it might have a situation in both x and y of the vehicle position are 0. Or it is also better if you could just set a boolean value instead of actual position value since we just want to know is there any data flow into the topic.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
```diff
@@ -122,23 +120,25 @@ def initialize():
...
+                    print("Error: Invalid selected route")
+                    rospy.spin()
                 
             
-    else:
-        print("Localization Incomplete")
-        rospy.spin()
+        else:
+            print("Localization Incomplete")
+            rospy.spin()
```
chengyuan0124, line 141: rospy.spin() should be removed to prevent the loop being stuck

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
```diff
@@ -122,23 +120,25 @@ def initialize():
...
-
+    while rospy.is_shutdown() != True:  
+        if local_pos.pose.position.x != 0 and local_pos.pose.position.y != 0:
+            if available_routes == None:
+                print("No routes available.")
             else:
-                print("Error: Invalid selected route")
-                rospy.spin()
+                """Check if the selected route is one of the available routes. If it isn't,then print error statement"""
+                if check_selected_route(available_routes) == True:
```
chengyuan0124, line 129: This line should be changed `if check_selected_route(available_routes):`

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_guidance
```diff
@@ -35,15 +35,14 @@ def initialize():
     rospy.init_node("carma_carla_guidance")
     service = rospy.ServiceProxy('/set_active_guidance')
```
chengyuan0124, line 38: Could you double check with this line of code a little bit? I am afraid that is not a acceptable usage. This is the usage I found from ROS python official website: [link](http://wiki.ros.org/ROS/Tutorials/WritingServiceClient%28python%29#:~:text=11%20%20%20%20%20try%3A%0A%20%2012-,add_two_ints%20%3D%20rospy.ServiceProxy(%27add_two_ints%27%2C%20AddTwoInts),-13%20%20%20%20%20%20%20%20%20resp1%20%3D%20add_two_ints)

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_guidance
```diff
@@ -35,15 +35,14 @@ def initialize():
...
-        route_sub = rospy.Subscriber("/route_event",RouteEvent(), check_route_status)
-        plugin_sub = rospy.Subscriber("/plugin_discovery", Plugin(), check_plugin_status)
-        if route_selected == True:
-            if plugins_activated == True:
+    
+    route_sub = rospy.Subscriber("/route_event",RouteEvent(), check_route_status)
+    plugin_sub = rospy.Subscriber("/plugin_discovery", Plugin(), check_plugin_status)
+    while rospy.is_shutdown() != True:  
+        if route_selected == True and plugins_activated == True:
                 service.call(True) #Sets guidance status to active
```
chengyuan0124: Same issue of the method usage

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
```diff
@@ -121,25 +116,20 @@ def initialize():
...
                     route = selected_route
                     set_route(route.route_id)
                     break
 
                 else:
-                    print("Error: Invalid selected route")
-                    rospy.spin()
-                
-            
+                    print("Error: Invalid selected route")
```
chengyuan0124, line 130: Are we going to keep loop running even the selected route is not one of the available routes?

carla-carma-integration/carla-carma-agent/agent/bridge.launch
```diff
@@ -93,4 +93,28 @@
     <param name='max_steering_degree' type="double" value="$(arg max_steering_degree)"/>
   </node>
 
+  <!--
+    # route #
+    Set the vehicle route after localization.
+  -->
+  <node pkg='carla_carma_bridge' type='carma_carla_route' name='carma_carla_route' output='screen'>
+    <param name='role_name' type="str" value="$(arg role_name)"/>
```
chengyuan0124, line 101: Not sure but do we need ego vehicle role name in `carma_carla_route`? Also, the `select_route` should be set here since there is a `get_param` method being called [here](https://github.com/usdot-fhwa-stol/carma-simulation/blob/81b8c891c1d440c383adcab1f1f5b4bec4a995b9/carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route#L32)

carla-carma-integration/carla-carma-agent/agent/bridge.launch
```diff
@@ -93,4 +93,28 @@
...
+  </node>
+
+  <!--
+    # plugins #
+    Activate all registered plugins.
+  -->
+  <node pkg='carla_carma_bridge' type='carma_carla_plugins' name='carma_carla_plugins' output='screen'>
+    <param name='role_name' type="str" value="$(arg role_name)"/>
```
chengyuan0124, line 109: Same question, do we need a role name?

carla-carma-integration/carla-carma-agent/agent/bridge.launch
```diff
@@ -93,4 +93,28 @@
...
+  </node>
+
+  <!--
+    # guidance #
+    Set guidance to active once the plugins have been activated and the route has been selected.
+  -->
+  <node pkg='carla_carma_bridge' type='carma_carla_guidance' name='carma_carla_guidance' output='screen'>
+    <param name='role_name' type="str" value="$(arg role_name)"/>
```
chengyuan0124, line 117: Same question, do we need a role name?

carla-carma-integration/carla-carma-agent/agent/bridge.launch
kjrush: Hmm, I guess that depends on if we're interacting with the CARLA ROSbridge topics right? If this node subscribes to the CARMA-side of those interfaces it shouldn't need the role name I think.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_guidance
kjrush: I think this comment is handled, right? Looks like the global call on line 25 should do what you've described.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_guidance
```diff
@@ -0,0 +1,45 @@
...
+def check_route_status(route_event_msg):
+    if route_event_msg.event == 1:
+        print("Route Selected")
+        route_selected = True
+
+def check_plugin_status(plugin_msg):
+    if plugin_msg.activated == True:
+        print("Plugins Activated")
+        #NOTE: May have to add logic to parse through every individual plugin and determine their status
+        plugins_activated = True
```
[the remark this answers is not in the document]
kjrush: Seems resolved.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_guidance
```diff
@@ -0,0 +1,52 @@
...
+
+def initialize():
+    rospy.init_node("carma_carla_guidance")
+    service = rospy.ServiceProxy('/set_active_guidance', SetGuidanceActive)
+    #While-loop monitors the route and plugin status and sets guidance to active once both are confirmed true
+    
+    route_sub = rospy.Subscriber("/route_event",RouteEvent(), check_route_status)
+    plugin_sub = rospy.Subscriber("/plugin_discovery", Plugin(), check_plugin_status)
+    while rospy.is_shutdown() != True:
```
kjrush, line 44: I think you need a spinOnce at the bottom of this while loop, otherwise you're not processing ROS messages at all and just doing a really tight loop here over variables that won't change.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_plugins
```diff
@@ -0,0 +1,79 @@
...
+    if plugin in plugin_list.plugins:
+        return True
+    else:
+        return False
+    
+
+def initialize():
+    rospy.init_node("carma_carla_plugins")
+    registered_plugins_serv = rospy.ServiceProxy('/get_registered_plugins',GetRegisteredPlugins)
+    plugin_list.plugins = registered_plugins_serv()
```
kjrush, line 72: I think you need a `global` on this plugin_list

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_plugins
```diff
@@ -0,0 +1,79 @@
...
+def plugin_status_cb(plugin_msg):
+    print("Received status from: " ++ plugin_msg.name)
+    requested = check_plugin_registration(plugin_msg)
+
+    #If the requested plugin is not registered, print error message
+    if requested != None:
+        print("Incoming plugin is not included in list of registered plugins")
+    elif check_selected_plugins(plugin_msg.name) == True:
+        plugin_msg.activated = True 
+        activation = activation_srv(plugin_msg.name, plugin_msg.activated)
```
kjrush, line 57: We should probably only be calling this once instead of every time we get a public status message. Not sure how you'd want to track that, but calling it every time we get one of these messages is a lot of unneeded calls. Maybe only call it if the status message we received indicates that the plugin is not activated?

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
kjrush: Any update on this?

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
```diff
@@ -0,0 +1,141 @@
...
+        return selected_route
+
+def initialize():
+    rospy.init_node("carma_carla_route")
+    route_event_sub = rospy.Subscriber('/guidance/route_event', RouteEvent,route_event_callback)
+    available_routes = get_available_routes()
+    
+    #Ensure that localization is complete before checking route information
+    localized_sub = rospy.Subscriber('/localization/current_pose', PoseStamped, pose_cb)
+    while rospy.is_shutdown() != True:
```
kjrush, line 118: Likewise I think you need a spinOnce call at the bottom of this while loop.

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_guidance
lewisid: [It seems like there's no spinOnce command for rospy. ](https://answers.ros.org/question/110336/python-spin-once-equivalent/)

carla-carma-integration/carla_carma_bridge/src/carla_carma_bridge/carma_carla_route
```diff
@@ -0,0 +1,140 @@
...
+        return available_routes.availableRoutes
+
+def set_route(id):
+    service = rospy.ServiceProxy('/set_active_route', SetActiveRoute)
+
+    try:
+        #Service Call Request
+        resp = service(0, id)
+        if resp.errorStatus != NO_ERROR.value:
+            if resp.errorStatus == NO_ROUTE.value:
```
kjrush, line 68: More what I was referring to was I don't understand the purpose of these internal if-statements and checks. You've removed the `errorDescription` variable (though it's referenced below in a few places so it still exists), but  as far as I can tell these checks on the error code don't have any function. I think it should be fine to just check `if there was an error: log/warn/etc. else: proceed with execution`

kjrush (approved):

PR #101, opened 2023-02-09
Merge release Isetta

SaikrishnaBairamoni:
# PR Details
This PR is to merge release branch isetta to master for release ISETTA
## Types of changes

## Checklist:


codygarver (dismissed):

docker/checkout.sh
```diff
@@ -39,6 +39,6 @@ if [[ "$BRANCH" = "develop" ]]; then
       git clone https://github.com/usdot-fhwa-stol/carma-msgs.git ~/src/carma-msgs --branch $BRANCH --depth 1
       git clone https://github.com/usdot-fhwa-stol/carma-utils.git ~/src/carma-utils --branch $BRANCH --depth 1    
 else
-      git clone https://github.com/usdot-fhwa-stol/carma-msgs.git ${dir}/src/carma-msgs --branch develop --depth 1
-      git clone https://github.com/usdot-fhwa-stol/carma-utils.git ${dir}/src/carma-utils --branch develop --depth 1
+      git clone https://github.com/usdot-fhwa-stol/carma-msgs.git ${dir}/src/carma-msgs --branch carma-system-4.3.0 --depth 1
```
kjrush, line 42: I think this should actually be 4.2.0, since these should match the version of CARMA Platform integrated with the system. @chengyuan0124 Can you confirm?

docker/install.sh
```diff
@@ -88,6 +88,7 @@ python3.7 -m pip install lxml==4.5.0
 
 # Install CARLA
 cd /home/carma/src/
+wget "https://carla-releases.s3.eu-west-3.amazonaws.com/Linux/CARLA_0.9.10.tar.gz"
```
kjrush, line 91: When was this added? I must have missed this PR somehow, I don't think this is correct. The whole next few lines are so that you can provide your own local copy of this file instead. Otherwise we're downloading over 6GB of this package every time we build. Functionally, it's correct. So if this is what we've been testing on, I'm not going to hold this up. But we should change this back on develop or we're really slowing things down.

kjrush (changes requested):

docker/checkout.sh
chengyuan0124: Yes, the testing we've done for XiL was based on 4.2.0 version

kjrush (approved):

PR #108, opened 2023-02-21
SUMO multi client feature for EVC integration

chengyuan0124:
# PR Details
## Types of changes

## Checklist:

### DIFF
--- a/co-simulation/bridge/carla_integration/sumo_simulation.py
+++ b/co-simulation/bridge/carla_integration/sumo_simulation.py
@@ -228,25 +228,42 @@ class SumoTLManager(object):
             signals.update(self._tls[tlid][program_id].get_associated_signals(landmark_id))
         return signals
 
-    def get_live_state(self, landmark_id):
+
+    def get_state_from_sumo(self, landmark_id):
+        """
+        This function will send a request to get sumo current traffic light state via MOSAIC
+        Returns the traffic light state of the signals associated with the given landmark.
+        """
         link_index_list = []
         _tlid = -1
         for tlid, link_index in self.get_all_associated_signals(landmark_id):
             _tlid = tlid
             link_index_list.append(link_index)
 
-        state_string = traci.trafficlight.getRedYellowGreenState(_tlid)
-        state_list = []
-        for link_index in link_index_list:
-            state_list.append(state_string[link_index])
-        for state in state_list:
-            if state == 'g' or state == 'G':
-                return SumoSignalState.GREEN_WITHOUT_PRIORITY
-            elif state == 'y':
-                return SumoSignalState.YELLOW
-            else:
-                return SumoSignalState.RED
+        ## SUMO traffic light state, ex 'rrggrr' in a intersection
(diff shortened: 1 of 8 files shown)


co-simulation/bridge/carla_integration/sumo_simulation.py
```diff
@@ -226,6 +228,26 @@ def get_all_associated_signals(self, landmark_id):
...
+        link_index_list = []
+        _tlid = -1
+        for tlid, link_index in self.get_all_associated_signals(landmark_id):
+            _tlid = tlid
+            link_index_list.append(link_index)
+
+        state_string = traci.trafficlight.getRedYellowGreenState(_tlid)
+        state_list = []
+        for link_index in link_index_list:
+            state_list.append(state_string[link_index])
```
kjrush, line 241: I think I've seen this before, but is this standard behavior for SUMO? I believe previously I've seen SUMO output traffic signal states as "gggGGGGyyyyR" or something like that, which I understand to be the state of each movement or link (whatever they're called) controlled by that signal is that correct?

I'm not a huge fan of using strings in place of more formal data types, but if that's what SUMO does, then I can live with it. I think this logic does effectively only check the first element in that list though right? This for loop will pull off the first element in the list, and it's going to fall into one of these branches with a return statement?

If it's intended to just look at the first element, you can just replace the for block with state_list[0], if you're actually intending to iterate through the whole list then we might need to rework this logic a bit.

co-simulation/bridge/carla_integration/synchronization.py
```diff
@@ -134,7 +137,13 @@ def tick(self):
                 carla_tl_state = BridgeHelper.get_carla_traffic_light_state(sumo_tl_state)
 
                 self.carla.synchronize_traffic_light(landmark_id, carla_tl_state)
+        elif self.tls_manager == 'EVC':
+            common_landmarks = self.sumo.traffic_light_ids & self.carla.traffic_light_ids
```
kjrush, line 141: Is & what you want here? I'm actually not 100% how bitwise AND applies to lists like this (or they seem to be lists?). I think you want `self.sumo.traffic_light_ids + self.carla.traffic_light_ids` if you're looking for the union of these sets.

Or does & in this context work to get only the elements that are the same in both lists? Is that what you want here?

chengyuan0124: `self.sumo.traffic_light_ids` and `self.carla.traffic_light_ids` are dictionaries that are similar in format to JSON. The line `self.sumo.traffic_light_ids & self.carla.traffic_light_ids` is used to obtain the intersection of the values in both dictionaries. This is done to verify that the `linkSignalID` value from CARLA has been properly reflected in the sumo `.net.xml` file to avoid any issues where a id cannot be found when trying to synchronize traffic light state from one simulation to another

co-simulation/bridge/carla_integration/sumo_simulation.py
chengyuan0124: That is correct. I've added comments in this function and modified the logic in the recent commit. Thanks!

PR #115, opened 2023-03-20
Implement Loop Detector Functionality in EVC-SUMO Bridge

chengyuan0124:
# PR Details
## Description

This pull request is adding the final piece to the EVC-SUMO bridge by implementing loop detector functionality. Specifically, the EVC-SUMO bridge will now be able to retrieve the loop detector status from the SUMO simulation object 'induction loop'. This object is created by using the SUMO tool netedit and defining the corresponding ID in the evc_sumo_cfg.json file. Once this information is obtained, the EVC-SUMO bridge will update the phase detector in EVC accordingly.

## Types of changes

## Checklist:


evc-sumo/src/evc_connector.py
```diff
@@ -88,23 +84,51 @@ def get_traffic_light_status_from_EVC(self, controller_io, phases):
...
-            state_num = state_num + len(phase["sumoTlStateIndex"])
-        ## init sumo tl state string with all red states
-        state_string = ['r'] * state_num
+        ## get number of characters in state string for SUMO and init sumo tl state string with all red
+        state_string = ['r'] * sum(len(phase['sumoTlStateIndex']) for phase in phases)
 
         for phase in phases:
             state_string = [self.COB_to_traffic_light_status(controller_io, phase['evcPhaseId']) if i in phase['sumoTlStateIndex'] else x for i,x in enumerate(state_string)]
-        return ''.join(state_string)
+        if None in state_string:
```
kjrush, line 96: I don't think None can be an element of a String right? A string variable could, itself be None, but I don't think you could have "rrrrryyNonegggg" or something. Or, after a quick google search you can but it would be that 
```
  a = "foo"
  b = None
  c = "bar
   
  out = a + b + c
  
```

out would be the string "fooNonebar" where None is simply text, not the actual None literal.

evc-sumo/src/evc_connector.py
```diff
@@ -88,23 +84,51 @@ def get_traffic_light_status_from_EVC(self, controller_io, phases):
...
+        if sumo_induction_loop_status != 0:
+            if controller_io.is_cib_off(evc_phase_id - 1):
+                controller_io.cib_on(evc_phase_id - 1)
+                return 1
+            else:
+                return 0
+        else:
+            if controller_io.is_cib_on(evc_phase_id - 1):
+                controller_io.cib_off(evc_phase_id - 1)
+                return 1
```
kjrush, line 129: This turns the CIB off but returns 1, should it be 0 instead?

evc-sumo/src/evc_connector.py
```diff
@@ -70,7 +70,8 @@ def COB_to_traffic_light_status(self, controller_io, evc_phase_id):
         elif controller_io.is_cob_on( (evc_phase_id - 1) + (MAX_INTERSECTION_PHASES * 0) ):
             return 'g'
         else:
-            return None
+            ## if state is not r, y or g then return O which is switched off state
+            return 'O'
```
kjrush, line 74: This makes sense, but I think there's a comment below that refers to None being the exception state, which is now '0'. I'd recommend correcting that comment to match the code and then I think we're good!

kjrush: Line 89, I think. Worth a quick check if there's anywhere else too

kjrush (approved):

PR #122, opened 2023-04-07
CARMA simulation and CARMA street handshake

chengyuan0124:
# PR Details
## Description

This pull request adds functionality to enable the handshake between CARMA simulation and CARMA street. The data flow of the handshake is as follows:

1. CARMA street sends an infrastructure registration message to mosaic-infrastructure.
2. The infrastructureRegistrationReceiver receives the registration message from CARMA street and converts it to an infrastructureRegistrationMessage object.
3. The infrastructureRegistrationReceiver stores the infrastructureRegistrationMessage object in a queue.
4. The infrastructureMessageAmbassador runs in a loop and checks the queue for any new messages while processing time advance grants.
5. When a new message is found in the queue, the infrastructureMessageAmbassador creates an RSU with its DSRC and sends the registration message to the MOSAIC RTI.
6. These created RSUs are stored in the InfrastructureInstanceManager with the type InfrastructureInstance.

## Types of changes

## Checklist:


co-simulation/fed/mosaic-carma/src/main/java/org/eclipse/mosaic/fed/carma/ambassador/CarmaRegistrationReceiver.java
```diff
@@ -64,7 +64,7 @@ public void run() {
             }
 
             // parse message
-            String receivedPacket = new String(msg.getData());
+            String receivedPacket = new String(msg.getData(), 0, msg.getLength());
```
kjrush, line 67: Was halfway through writing up a question as to why you chose to use this version of the constructor before I realized the error you were correcting, very nice catch! Previously it was encoding the whole buf[] contents even if the received packet was smaller than `UDP_MTU` bytes right?

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
```diff
@@ -25,20 +25,52 @@
...
  */
 public class InfrastructureInstanceManager {
-    private Map<String, InfrastructureInstance>  managedInstances = new HashMap<>();
+    private Map<String, InfrastructureInstance> managedInstances = new HashMap<>();
     private double currentSimulationTime;
- 
+
+    public void onNewRegistration(InfrastructureRegistrationMessage registration){
+        if (!managedInstances.containsKey(registration.getInfrastructureId())) {
+            try {
```
kjrush, line 41: I think this try/catch block is redundant with the one in `newInfrastructureInstance` and can be removed. I believe UnknownHostException is a sub-class of IOException anyway, so the block around `tmp.bind()` would catch it, convert it into a RuntimeException which this block wouldn't catch anyway.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
```diff
@@ -25,20 +25,52 @@
...
+            // log warning
+        }
+    }
+
+    private void newInfrastructureInstance(String infrastructureId, InetAddress rxMessageIpAddress, int rxMessagePort, int timeSyncPort, GeoPoint location) {
+        InfrastructureInstance tmp = new InfrastructureInstance(infrastructureId, rxMessageIpAddress, rxMessagePort, timeSyncPort, location);
+        try {
+            tmp.bind();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
kjrush, line 62: Worth also logging some sort of information here as well.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
```diff
@@ -25,20 +25,52 @@
...
+        }
+        managedInstances.put(infrastructureId, tmp);
+    }
+
+        /**
```
kjrush, line 67: Weird indentation here, just fyi.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureMessageAmbassador.java
```diff
@@ -47,21 +46,16 @@ public class InfrastructureMessageAmbassador extends AbstractFederateAmbassador
...
 
-    /**
-     * List of infrastructures that are controlled
-     */
-    private final HashMap<String, Boolean> v2xMap = new HashMap<>();
-
     /**
      * The number of CARMA vehicles.
      */
     int numberOfInfrastructureInstances = 0;
```
kjrush, line 52: Not part of your changeset, but this should almost certainly be private. And the comment above references "CARMA Vehicles" should be changed to infrastructure

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureMessageAmbassador.java
```diff
@@ -149,44 +138,120 @@ private synchronized void receiveInteraction(InfrastructureV2xMessageReception i
...
+    }
 
+    /**
+     * Creates an Ad-Hoc configuration object to represent the configuration of the
+     * Ad-Hoc interface used for communication between the infrastructure instance
+     * and other vehicles or components, and sends it to the RTI for exchange.
+     *
+     * @param reg the infrastructure registration message received from the RTI
+     */
+    private final void onDsrcRegistrationRequest(InfrastructureRegistrationMessage reg) {
```
kjrush, line 150: Declaring this "final" is probably unnecessary, since we're unlikely to subclass this. Honestly probably better to explicitly declare things final more often, but it seems a bit inconsistent with the rest of what's currently written.

Also, it's super super minor, but something about passing the whole InfrastructureRegistrationMessage as the input here seems a bit off to me. I think slightly better would be either passing just the infrastructure ID (since that's all that's needed, data-wise) or passing in the actual InfrastructureInstance or something. As it is now, you could call this method prior to the `onRsuRegistration...` method below which I think would result in NS-3 not properly processing the registration. You could lean a bit more on the type-system here (by having the parameter here be of a type created by the `onRsuRegistration...` method to more clearly express the ordering just through the data types. Having them be the same input type in this case almost implies they work independently rather than sequentially. 

I don't think you need to make any changes here, but just wanted to share something to consider for future stuff.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureMessageAmbassador.java
```diff
@@ -149,44 +138,120 @@ private synchronized void receiveInteraction(InfrastructureV2xMessageReception i
...
+        // Create an InterfaceConfiguration object to represent the configuration of the
+        // Ad-Hoc interface
+        // Set the IP address and subnet mask to null for now
+        // Set the transmit power to 50 dBm and the maximum range to 100 meters
+        InterfaceConfiguration interfaceConfig = new InterfaceConfiguration.Builder(AdHocChannel.SCH1)
+                .ip(null)
+                .subnet(null)
+                .power(50)
```
kjrush, line 158: These are good for now, might be worth adding a TODO comment here for us to get back to making these parameters sometime, or writing up a JIRA ticket or something.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureMessageAmbassador.java
```diff
@@ -149,44 +138,120 @@ private synchronized void receiveInteraction(InfrastructureV2xMessageReception i
...
      *
-     * @param time The timestamp towards which the federate can advance it local
-     *             time.
+     * @param reg The registration message of the new infrastructure instance.
+     */
+    private final void onRsuRegistrationRequest(InfrastructureRegistrationMessage reg) {
+
+        // Register the new infrastructure instance to the RTI as an RSU
+        RsuRegistration rsuRegistration = new RsuRegistration(currentSimulationTime, reg.getInfrastructureId(),
+                null, null,
```
kjrush, line 194: `null` might be a problem unless you've tested this already. I could see the MOSAIC RTI choking if passed null here, unless they're on top of their null-checks. I think if we really want to say nothing for the group we should just use the empty string "", and for the applications I'd go with an empty-list.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureMessageAmbassador.java
kjrush: Might be worth adding to the doc comment that this should be called **after** the `onRsuRegistration...` method.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureRegistrationMessage.java
```diff
@@ -23,12 +23,67 @@
  */
 public class InfrastructureRegistrationMessage {
     private String rxMessageIpAddress;
-    private int timeSyncListenPort = 1517; // TODO
-    private int regstratinoListenPort = 1516; // TODO 
+    private String infrastructureId;
+
+    private int rxMessagePort = 1536;
+    private int timeSyncPort = 1517; // TODO
```
kjrush, line 29: TODO what? It's good to note what the todo-item is so that somebody else might pick it up later, though generally I like to be careful about cleaning up TODO items before a PR, anything items that remain at that point really need JIRA stories or github issues as they've become a separate work item.

kjrush (changes requested):

co-simulation/fed/mosaic-carma/src/main/java/org/eclipse/mosaic/fed/carma/ambassador/CarmaRegistrationReceiver.java
chengyuan0124: I found this issue when I was running the `InfrastructurethereRegistrationReceiverTest.java` which used same way to getting the packet and convert that to the specified class format. 
When we created the `String` object from the received bytes, it may contain extra characters at the end that were not part of the original message. This is because the `buf` array used in the `DatagramPacket` constructor is preallocated and may contain unused bytes. Therefore, the String object created from `msg.getData()` may have trailing null characters. 
This null characters causes an exception when `gson.fromJson` trying to convert received packet into the specified class format. So I modified the `String` object using the `String` constructor that takes a byte array and the length of the message to ensure that only the actual message bytes are used to create the `String`

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
chengyuan0124: This might be needed for the line `InetAddress.getByName(registration.getRxMessageIpAddress())` to see if the IP address is not valid or could not be resolved

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureMessageAmbassador.java
chengyuan0124: I believe this one can be removed since the list of instances can be accessed by using the function Eric implemented `getManagedInstances`

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
paulbourelly999: Not sure I understand catching and exception to throw a different exception, especially if we are not catching the new exception we throw. I would prefer to log the exception content with information about how we failed to register a new instance.

paulbourelly999 (approved):
Mostly looks good other than the small exception stuff.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureMessageAmbassador.java
kjrush: Yeah, I think i'd be okay with removing it if you want to go ahead and delete it in this PR, provided it's not used anywhere else.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
kjrush: Ah, okay, that makes more sense then

kjrush (commented):
I think generally looks good to me, if you can address Pauls comment on the exception logging and update the copyright statement on the InfrastructureMessageAmbassador to the correct Leidos one then I think we're good to go.

kjrush (approved):

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
```diff
@@ -25,20 +25,52 @@
...
+            try {
+                newInfrastructureInstance(
+                    registration.getInfrastructureId(),
+                    InetAddress.getByName(registration.getRxMessageIpAddress()),
+                    registration.getRxMessagePort(),
+                    registration.getTimeSyncPort(),
+                    registration.getLocation()
+                );
+            } catch (UnknownHostException e) {
+                throw new RuntimeException(e);
```
paulbourelly999: Similar comment about throwing new exception. Either catch exception and log and/or do something here. Or don't catch the exception here and catch it somewhere else and log/do something there.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
```diff
@@ -25,20 +25,52 @@
...
+            }
+        } else {
+            // log warning
```
paulbourelly999: Add log statement

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
chengyuan0124: This is modified in this commit: https://github.com/usdot-fhwa-stol/carma-simulation/pull/122/commits/d7b4c932ed70a0faf05d4209dfbf073e26cc956a

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
chengyuan0124: This is modified in this commit: https://github.com/usdot-fhwa-stol/carma-simulation/pull/122/commits/d7b4c932ed70a0faf05d4209dfbf073e26cc956a

paulbourelly999 (approved):
Looks good

PR #128, opened 2023-04-20
Fix/simulation frozen issue

chengyuan0124:
# PR Details
## Description

This PR fixes the simulation freezing issue that occurs when NS3 and infrastructure ambassador are enabled and running in the MOSAIC scenario Tiergarten. The solution involved testing a simple Python script that sends a fake infrastructure registration JSON message to the infrastructure message receiver. The test confirmed that the infrastructure ambassador successfully registered RSU and DSRC to NS3 and stored the new registration in the infrastructureInstanceManager.

This screenshot is the log retrieving from MOSAIC-infrastructure ambassador:

Screenshot from 2023-04-20 03-53-16

## Types of changes

## Checklist:


co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureMessageAmbassador.java
```diff
@@ -162,16 +164,17 @@ private synchronized void receiveInteraction(InfrastructureV2xMessageReception i
...
+    private void onDsrcRegistrationRequest(String infrastructureId) throws UnknownHostException {
         // Create an InterfaceConfiguration object to represent the configuration of the
         // Ad-Hoc interface
         // Set the IP address and subnet mask to null for now
         // Set the transmit power to 50 dBm and the maximum range to 100 meters
         // NOTE: TODO Setup the IP address and subnet
         InterfaceConfiguration interfaceConfig = new InterfaceConfiguration.Builder(AdHocChannel.SCH1)
-                .ip(null)
-                .subnet(null)
+                .ip( (Inet4Address) Inet4Address.getByName("192.168.0.1") ) //TODO
```
kjrush, line 176: Can you explain what the TODO here is? Generally leaving notes for what needs to be done is good, in case someone else ends up picking up where you left off so they have context.

kjrush (commented):
Looks fine to me, though I'm not sure exactly how this resolves the issue, or what the original issue was? Can you explain what the freezing was caused by and how this addresses it?

Also, the test case you described does that match how the issue was originally discovered? If not, it'd be good to repeat that exact test to fully verify that this fixes the particular issue that we experienced there.

kjrush:
Also, took a look at the sonar scanner report. We can ignore the security issue they're talking about entirely, it's not an issue since it's not an actual-literal-real IP address, just a string representing an address within our simulated communications network.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
```diff
@@ -94,6 +93,7 @@ private void newInfrastructureInstance(String infrastructureId, InetAddress rxMe
                 timeSyncPort, location);
         try {
             tmp.bind();
+            log.info("New infrastructure registration received with ID '{}'", infrastructureId);
```
paulbourelly999, line 96: This is a sort of deceiving log statement. To me, it suggests that the CARMASimulation instance just received a new registration message but really this is only triggered after a registration message has been received and the next time step is executed. Better wording maybe "New Infrastructure instance {} registered with Infrastructure Instance Manager." or something. This way we can add a log statement to the actual InfrastructureRegistrationReceiver to log when the registration is received and then separately log when we register the instance. Does that make sense? Eric and I spent some time trying to figure out why we could not trigger this log statement.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureRegistrationReceiver.java
```diff
@@ -38,9 +38,9 @@
 public class InfrastructureRegistrationReceiver implements Runnable {
     private Queue<InfrastructureRegistrationMessage> rxQueue = new LinkedList<>();
     private DatagramSocket listenSocket = null;
-    private static final int listenPort = 1515;
+    private static final int listenPort = 1615;
```
paulbourelly999, line 41: Why are we changing this port?

paulbourelly999: Also can we add a log statement to this class when a registration message is received?

chengyuan0124: port being change is because it is used by [CARMARegistrationReceiver](https://github.com/usdot-fhwa-stol/carma-simulation/blob/064633572e6ce9fe08747242b9d9765ca8ad3645/co-simulation/fed/mosaic-carma/src/main/java/org/eclipse/mosaic/fed/carma/ambassador/CarmaRegistrationReceiver.java#L39)

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureMessageAmbassador.java
```diff
@@ -162,16 +164,16 @@ private synchronized void receiveInteraction(InfrastructureV2xMessageReception i
...
+    private void onDsrcRegistrationRequest(String infrastructureId) throws UnknownHostException {
         // Create an InterfaceConfiguration object to represent the configuration of the
         // Ad-Hoc interface
         // Set the IP address and subnet mask to null for now
         // Set the transmit power to 50 dBm and the maximum range to 100 meters
-        // NOTE: TODO Setup the IP address and subnet
         InterfaceConfiguration interfaceConfig = new InterfaceConfiguration.Builder(AdHocChannel.SCH1)
-                .ip(null)
-                .subnet(null)
+                .ip( (Inet4Address) Inet4Address.getByName("192.168.0.1") ) //TODO: set IP address if needed
```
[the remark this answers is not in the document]
chengyuan0124: When a new infrastructure registration message is received, the DsrcRegistrationRequest is triggered. Currently, this parameter is hard-coded because our need does not allow for its configuration. However, it is necessary for the DSRC initialize function.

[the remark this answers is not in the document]
kjrush: I think this is an easy ignore on the sonar scanner, this is a virtual IP (e.g. one within the simulated context) not an actual network IP. I can't yet tell exactly how this IP is used within the context (is it the self-assigned IP address of the radio?) but it shouldn't matter as DRSC isn't an IP protocol anyway, so this wouldn't be used even if we had better data for it.

co-simulation/bundle/src/assembly/resources/scenarios/Tiergarten/infrastructure/infrastructure_config.json
```diff
@@ -0,0 +1,4 @@
+{
+    "updateInterval": 1000,
```
paulbourelly999, line 2: Last question what is update interval representing and what are its units. Sometimes it's value is 1000 and sometimes its 100. Is that intentional?

co-simulation/bundle/src/assembly/resources/scenarios/Town04/carla/carla_config.json
```diff
@@ -0,0 +1,6 @@
+{
+    "updateInterval": 100,
```
paulbourelly999, line 2: Why is update interface 100 here but 1000 above

paulbourelly999 (dismissed):
Looks good mostly. Just had that last question I included.

co-simulation/bundle/src/assembly/resources/scenarios/Town04/carma/carma_config.json
```diff
@@ -0,0 +1,25 @@
+{
+     "updateInterval": 100,
+    "carmaVehicles":[
+{   
+    "routeID": "0",
+    "lane": 1,
+    "position": 0,
+    "departSpeed": 0,
+    "vehicleType": "vehicle.chevrolet.impala",
+    "applications":["org.eclipse.mosaic.app.tutorial.VehicleCommunicationApp"],
```
kjrush, line 10: Is this usually part of the scenario? I think this is running the same thing that ODU had running, e.g. a basic MOSAIC Vehicle not a CARMA Platform vehicle right?

kjrush (approved):

paulbourelly999 (approved):

kjrush (approved):

PR #152, opened 2023-07-27
Add Sensor/Detector Registration to Infrastructure Ambassador

paulbourelly999:
# PR Details
## Description

Added additional information to the Infrastructure Registration Message including : 
- Port for transmitting relevant interactions
- List of sensors/detectors to register for infrastructure instance

Added functionality to process new registration message, trigger sensors/detectors registration interactions to prompt creation of sensors/detectors.
Added functionality to process incoming Detected Object interactions from sensors and forward them via the new interactions port to registered infrastructure instances based on the sensor that reports the detection.

## Motivation and Context

New functionality will be used to create sensor in CARLA and retrieve detection information for the VRU use case
## How Has This Been Tested?

Unit tested and Integration testing with XIL Cloud setup
Only tested that CDASim properly processes Infrastructure registration and still forwards time sync messages. Further testing for  detection interaction processing is advised once CARLA Ambassador work completes.
## Types of changes

## Checklist:

- [] I have updated the documentation accordingly.


co-simulation/fed/mosaic-infrastructure/pom.xml
```diff
@@ -25,6 +25,13 @@
       <artifactId>mosaic-objects</artifactId>
       <version>${mosaic.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.eclipse.mosaic</groupId>
+      <artifactId>mosaic-objects</artifactId>
+      <version>${mosaic.version}</version>
+      <type>test-jar</type>
```
paulbourelly999, line 32: This package is already included in the infrastructure ambassador pom as a dependency but I had to add the test-jar to gain access to Rules used to initialize singletons for unit testing. See GeoProjectionRule and IpResolverRule in the InfrastructureMessageAdapterTest.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java
```diff
@@ -156,7 +223,7 @@ public void setTimeSyncPort(int timeSyncPort) {
      *                     methods
      */
     public void bind() throws IOException {
-        rxMsgsSocket = new DatagramSocket();
+        socket = new DatagramSocket();
```
paulbourelly999, line 226: renamed to `socket` since this Datagramsocket is used now to transmit Time Sync, V2X and Simulated Interactions. RxMsgsSocket indicates it relates the only to the rxMessagePort used for V2X messages.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java
```diff
@@ -167,27 +234,40 @@ public void bind() throws IOException {
...
+        sendPacket(data, simulatedInteractionPort);
+    }
+    /**
+     * Method to send byte data to specified port using the infrastructure instance Datagramsocket.
+     * 
+     * @param data byte data to send using Datagramsocket.
+     * @param port in integer format to send Datagram to.
+     * @throws IOException
+     */
+    public void sendPacket(byte[] data, int port) throws IOException {
```
paulbourelly999, line 266: New method created to avoid code duplication between sendTimeSyncMsg sendV2xMsg and sendInteraction

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureRegistrationMessage.java
```diff
@@ -13,169 +13,161 @@
...
     /**
-     * Sets the unique identifier for the Infrastructure Device
-     * It currently is the same with RSU ID
+     * Returns port where infrastructure instance will be listening for interaction messages
+     * from CDASim.
      * 
-     * @param infrastructureId The unique identifier for the Infrastructure Device
+     * @return int simulated interaction port
      */
-    public void setInfrastructureId(String infrastructureId) {
```
paulbourelly999, line 134: Removed setters since they were unused and just require additional unit test code coverage.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureRegistrationReceiver.java
```diff
@@ -44,7 +44,7 @@ public class InfrastructureRegistrationReceiver implements Runnable {
     private Queue<InfrastructureRegistrationMessage> rxQueue = new LinkedList<>();
     private DatagramSocket listenSocket = null;
     private static final int listenPort = 1615;
-    private boolean running = false;
+    private AtomicBoolean running = new AtomicBoolean(false);
```
paulbourelly999, line 47: Since this is a runnable in which stop can be called outside the running thread, the running boolean flag should be thread safe against concurrent modification.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureRegistrationReceiver.java
```diff
@@ -70,13 +70,14 @@ public void init() {
...
         byte[] buf = new byte[UDP_MTU];
-        running = true;
-        while (running) {
+        running.set(true);
+        while (running.get()) {
             DatagramPacket msg = new DatagramPacket(buf, buf.length);
             try {
                 listenSocket.receive(msg);
             } catch (IOException e) {
-                throw new RuntimeException(e);
```
paulbourelly999, line 79: Throwing/Rethrowing this exception has the following issues. On each call to close, a SocketException will be triggered here since the `close` method is called outside the `run` while loop and there is currently no timeout configured and no sleep/periodic polling setup on the datagramsocket. This means that when stop is called and the socket is closed, the InfrastructureRegistrationReceiver thread is still blocked on the receive() call and then throws an exception because we have now closed the socket. The way I avoid this is on exception I continue and re-evaluate the running condition of the socket.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureTimeInterface.java
```diff
@@ -1,69 +0,0 @@
-/*
- * Copyright (C) 2023 LEIDOS.
- *
- * Licensed under the Apache License, Version 2.0 (the "License"); you may not
```
paulbourelly999, line 4: This class was not really necessary. It only contained really a single functionality and required the infrastructuremanager to do it. I instead moved this functionality to the infrastructuremanager

co-simulation/lib/mosaic-carma-utils/src/main/java/gov/dot/fhwa/saxton/CarmaV2xMessageReceiver.java
```diff
@@ -41,7 +42,7 @@ public class CarmaV2xMessageReceiver implements Runnable {
     private Queue<Tuple<InetAddress, CarmaV2xMessage>> rxQueue = new LinkedList<>();
     private DatagramSocket listenSocket = null;
     private int listenPort;
-    private boolean running = true;
+    private AtomicBoolean running = new AtomicBoolean(false);
```
paulbourelly999, line 45: Same comment as with the InfrastructureMessageReceiver.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java
```diff
@@ -131,22 +141,79 @@ public void setRxMessagePort(int rxMessagePort) {
...
+    }
+
+
+    /**
+     * Sets the simulated interaction port of the infrastructure instance.
+     * 
+     * @param simulatedInteractionPort int simulated interaction port of the infrastructure
+     * instance.
+     */
+    public void setSimulatedInteractionPort(int simulatedInteractionPort) {
```
dan-du-car, line 177: Maybe check if input to make sure it doesn't receive negative values. The same comment for other setter functions to update port numbers if use int data type.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java
```diff
@@ -167,27 +234,40 @@ public void bind() throws IOException {
...
+    }
+    /**
+     * Method to send byte data to specified port using the infrastructure instance Datagramsocket.
+     * 
+     * @param data byte data to send using Datagramsocket.
+     * @param port in integer format to send Datagram to.
+     * @throws IOException
+     */
+    public void sendPacket(byte[] data, int port) throws IOException {
+         if (socket == null) {
```
dan-du-car, line 267: https://docs.oracle.com/javase/7/docs/api/java/net/SocketException.html Do we know the specific socket exception at this point? Maybe checking whether the socket is ready when we initialize the infrastructure instance? so it guarantees that the socket is ready for use here?

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java
paulbourelly999: The more we change these methods, the more code coverage they require. The setter is actually never used anywhere in the code so I do not think it is worth the extra effort here. We also do not do this validation on the infrastructure registration so doing it here may not make much sense

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java
paulbourelly999: The only precondition for the socket is that it is not null/initialized by the bind method, which we already check. Additionally I am not sure editing the UDPSockets we use for CDASim Infrastructure Ambassador really falls under the scope of this story.

co-simulation/fed/mosaic-infrastructure/pom.xml
```diff
@@ -25,6 +25,13 @@
...
+      <scope>test</scope>
+    </dependency>
     <dependency>
         <groupId>gov.dot.fhwa.saxton</groupId>
         <artifactId>mosaic-carma-utils</artifactId>
```
dan-du-car, line 37: Can we make this mosaic-carma-utils artifact group id consistent as the other artifacts in the lib? Eg: mosaic-objects is in group org.eclipse.mosaic, but the mosaic-carma-utils is in group gov.dot.fhwa.saxton?

co-simulation/lib/mosaic-carma-utils/src/main/java/gov/dot/fhwa/saxton/CarmaV2xMessageReceiver.java
```diff
@@ -105,18 +108,20 @@ public void run() {
            } catch (IllegalArgumentException parseError) {
                log.warn("CarmaV2xMessageReceiver received malformed message with length: " + msg.getLength() + " from client " + msg.getAddress().toString() + "! Reason: " + parseError.getMessage() +". Discarding...");
            }
-       }
+        }
+        if (listenSocket != null) {
```
dan-du-car, line 112: This is redundant call. If it is not running, it must be called to stop(). The socket is release in the stop() function already.

co-simulation/fed/mosaic-infrastructure/pom.xml
paulbourelly999: No mosiac-carma-utils is a jar we created while mosiac-objects is a jar created by eclipse mosiac. This may mean that I need to move my detection objects into a mosiac-carma-objects library.

co-simulation/lib/mosaic-carma-utils/src/main/java/gov/dot/fhwa/saxton/CarmaV2xMessageReceiver.java
paulbourelly999: Will address

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/DetectedObject.java
```diff
@@ -0,0 +1,249 @@
...
+import org.eclipse.mosaic.lib.geo.CartesianPoint;
+import org.eclipse.mosaic.lib.math.Vector3d;
+
+public final class DetectedObject implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    private DetectionType type;
+
+    private double confidence;
```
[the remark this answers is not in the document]
paulbourelly999: Will address

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
```diff
@@ -91,25 +101,30 @@ public void onNewRegistration(InfrastructureRegistrationMessage registration) {
      * 
      */
     private void newInfrastructureInstance(String infrastructureId, InetAddress rxMessageIpAddress, int rxMessagePort,
```
dan-du-car, line 103: This function name is a bit misleading. It looks like trying to create a new object. The name does not account for adding the object into a global map of instances.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
```diff
@@ -91,25 +101,30 @@ public void onNewRegistration(InfrastructureRegistrationMessage registration) {
...
+
         managedInstances.put(infrastructureId, tmp);
     }
 
     /**
-     * Callback to be invoked when CARMA Platform receives a V2X Message from the NS-3 simulation
+     * Callback to be invoked when CARMA Platform receives a V2X Message from the
+     * NS-3 simulation
+     * 
      * @param sourceAddr The V2X Message received
```
dan-du-car, line 124: The V2X Message received => The source address where the V2X Message is from

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
```diff
@@ -91,25 +101,30 @@ public void onNewRegistration(InfrastructureRegistrationMessage registration) {
...
-     * @param txMsg The Host ID of the vehicle receiving the data
-     * @throws RuntimeException If the socket used to communicate with the platform experiences failure
+     * @param txMsg      The Host ID of the vehicle receiving the data
```
dan-du-car, line 125: The actual V2x message that received from platform

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
```diff
@@ -91,25 +101,30 @@ public void onNewRegistration(InfrastructureRegistrationMessage registration) {
...
+     * @throws RuntimeException If the socket used to communicate with the platform
+     *                          experiences failure
      */
     public V2xMessageTransmission onV2XMessageTx(InetAddress sourceAddr, CarmaV2xMessage txMsg, long time) {
```
dan-du-car, line 129: "onV2xMessageTx" What does the Tx stands for?

dan-du-car: There is a time parameter. What is unit of measure for this time? What is purpose of this time in this function?

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
```diff
@@ -135,24 +151,99 @@ public V2xMessageTransmission onV2XMessageTx(InetAddress sourceAddr, CarmaV2xMes
...
     public void onV2XMessageRx(byte[] rxMsg, String rxRsuId) {
-        if (!managedInstances.containsKey(rxRsuId))  {
+        if (!managedInstances.containsKey(rxRsuId)) {
             return;
         }
 
         InfrastructureInstance rsu = managedInstances.get(rxRsuId);
         try {
-            rsu.sendMsgs(rxMsg);
+            rsu.sendV2xMsg(rxMsg);
```
dan-du-car, line 169: This callback says receive message, but it also send V2x message to RSU. Maybe update the brief to include sending the received  V2X message to RSU?

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
```diff
@@ -135,24 +151,99 @@ public V2xMessageTransmission onV2XMessageTx(InetAddress sourceAddr, CarmaV2xMes
     }
 
     /**
-     * Callback to be invoked when CARMA Platform receives a V2X Message from the NS-3 simulation
-     * @param rxMsg The V2X Message received
+     * Callback to be invoked when an RSU receives a V2X Message from the NS-3
+     * simulation
+     * 
+     * @param rxMsg   The V2X Message received
      * @param rxRsuId The Host ID of the vehicle receiving the data
```
dan-du-car, line 158: rxRsuId comment is not clear

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
```diff
@@ -135,24 +151,99 @@ public V2xMessageTransmission onV2XMessageTx(InetAddress sourceAddr, CarmaV2xMes
...
+    /**
+     * Callback to be invoked when a infrastructure instance receives a simulated
+     * object detection from a registered simulated sensors. Only send object detection
+     * to the infrastructure instance that contains the sensor that reported it.
+     * 
+     * @param detection Detected Object.
+     */
+    public void onDetectedObject(DetectedObject detection) {
+        for (InfrastructureInstance instance : managedInstances.values()) {
+            if (instance.containsSensor(detection.getSensorId())) {
```
dan-du-car, line 184: Do we have to identify the infrastructure instance registered with this particular sensor ID every time we are trying to forward detected object? Can we map the Detector to the sensor ID, so that we can directly access map with the sensor ID key. I think it might be faster to use Java standard library instead of implementing our own lookup with two nested loop if this lookup is frequently used as we constantly sending detected object?

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
```diff
@@ -135,24 +151,99 @@ public V2xMessageTransmission onV2XMessageTx(InetAddress sourceAddr, CarmaV2xMes
...
+    }
+
+    /**
+     * Helper method to serialize time sychronization message to JSON and encode into bytes 
+     * for eventual transmission via a Datagramsocket.
+     * 
+     * @param message time synchronization message to encode
+     * @return resulting bytes encoded from JSON string.
+     */
+    private byte[] encodeTimeMessage(InfrastructureTimeMessage message) {
```
dan-du-car, line 204: Feel like we can combine encodeTimeMessage and asJon into one function such as encodeMessage. Then replace all encode* functions with the generic encodeMessage function.

co-simulation/fed/mosaic-infrastructure/pom.xml
dan-du-car: I do not think we need to move the detection object into library. This is just group ID changes to move `mosaic-carma-utils` into `org.eclipse.mosaic` group. And replace the namespace `gov.dot.fhwa.saxton` in the `mosaic-carma-utils` to `org.eclipse.mosaic`

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/DetectorType.java
```diff
@@ -0,0 +1,63 @@
...
+    
+    /**
+     * Returns the enum mapped from an String name.
+     *
+     * @param label string.
+     * @return the enum mapped from String name.
+     */
+    public static DetectorType fromLabel(String label) {
+        for (DetectorType type: DetectorType.values()) {
+            if (type.label.equals(label)) {
```
dan-du-car, line 53: Maybe case insensitive comparison and remove any hidden leading or tailing spaces.

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/Orientation.java
```diff
@@ -0,0 +1,83 @@
...
+ * the License.
+ */
+package org.eclipse.mosaic.lib.objects.detector;
+
+import java.io.Serializable;
+
+public class Orientation implements Serializable{
+    private static final long serialVersionUID = 1L;
+
+    private double yaw;
```
dan-du-car, line 23: JavaDoc comment for each member

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/Size.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+package org.eclipse.mosaic.lib.objects.detector;
+
+import java.io.Serializable;
+
+public class Size implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    private double length;
+
```
dan-du-car, line 24: JavaDoc comment for each member

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/gson/DetectorTypeAdapter.java
```diff
@@ -0,0 +1,42 @@
...
+import org.eclipse.mosaic.lib.objects.detector.DetectorType;
+
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonParseException;
+import com.google.gson.JsonSerializationContext;
+import com.google.gson.JsonSerializer;
+
+public class DetectorTypeAdapter implements JsonSerializer<DetectorType>, JsonDeserializer<DetectorType> {
```
dan-du-car, line 29: JavaDoc comments describe the purpose of this class

dan-du-car: Maybe rename it with DetectorTypeSerializer?

co-simulation/lib/mosaic-objects/src/test/java/org/eclipse/mosaic/lib/objects/detector/DetectorTypeTest.java
```diff
@@ -0,0 +1,34 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
+public class DetectorTypeTest {
+    @Test
+    public void testFromLabel() {
+        DetectorType type = DetectorType.fromLabel("SemanticLidar");
```
dan-du-car, line 27: Do we have a unit test for detectorTypeAdapter?

co-simulation/fed/mosaic-infrastructure/pom.xml
dan-du-car: This is the group id I refer to: https://github.com/usdot-fhwa-stol/cdasim/blob/35c45a53d499c7ba8901177ffd73f5b6a0d32fab/co-simulation/lib/mosaic-carma-utils/pom.xml#L13.
The package is here: https://github.com/usdot-fhwa-stol/cdasim/blob/35c45a53d499c7ba8901177ffd73f5b6a0d32fab/co-simulation/lib/mosaic-carma-utils/src/main/java/gov/dot/fhwa/saxton/CarmaV2xMessage.java#L17

https://github.com/usdot-fhwa-stol/cdasim/blob/35c45a53d499c7ba8901177ffd73f5b6a0d32fab/co-simulation/lib/mosaic-carma-utils/src/main/java/gov/dot/fhwa/saxton/CarmaV2xMessageReceiver.java#L17

paulbourelly999: I am aware of the group id. That is mean to indicate the organization responsible for creating the code. Mosiac is an Eclipse managed open source simulation framework we developed on top of.  The packages we developed are developed under the package structure gov.dot.fhwa.saxton to indicate they are stol packages while the mosiac packages are developed under the org.eclipse.mosiac.lib package structure

paulbourelly999: https://github.com/eclipse/mosaic

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java
```diff
@@ -35,40 +37,48 @@ public class InfrastructureInstance {
     private InetAddress targetAddress;
     private int rxMessagePort;
     private int timeSyncPort;
+    private int simulatedInteractionPort;
```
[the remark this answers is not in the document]
paulbourelly999: I think that is probably overkill, especially since all the getter/setter methods have documentation. I can add documentation to the constructor as well but I do not think it is necessary for every class member

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
paulbourelly999: This method is not part of my PR and does exactly what it says it does. This is an Infrastructure Instance Manager. A class that managers Infrastructure Instances. This method adds an infrastructure instance to the managed instances. U di bit think it is misleading at all. I could add more documentation to the method but even that is kind of outside my PR since my PR is about adding sensor registration and object detection.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
paulbourelly999: I can edited this method comment as well but it is also not part of my PR. Looks like some formatting changes changed the comment format but there are not actual changes to the method.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
paulbourelly999: I can edited this method comment as well but it is also not part of my PR. Looks like some formatting changes changed the comment format but there are not actual changes to the method.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
paulbourelly999: Rx and Tx universally mean transmit and receive. The time refers to simulation time, specifically the time for the transmission interaction. This method is also not part of my PR. I think this review should focus on code changes.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
paulbourelly999: receiving RSU ID. Also not part of my PR. Just formatting differences

co-simulation/fed/mosaic-infrastructure/pom.xml
dan-du-car: I see. It is a bit confusing, if it is a clone a another repos, not sure why we made a full copy here instead of just using a clone reference to that repos. This mosaic-carma-utils package is located under this lib but has a completely out of context package name `gov.dot.fhwa.saxton` from my point of view. Initially, I thought this mosaic-carma-utils is from another repos because of the package name `gov.dot.fhwa.saxton`.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
paulbourelly999: The receptions refers to a V2X message receptions interaction. NS3 has triggered an interaction that indicates that the given RSU should receive a given V2X message. When this RSU receives this message, it is forwarded to V2X-Hub.

paulbourelly999: Also not part of my PR. Just renamed them method.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
paulbourelly999: We can create a duplicate MAP for sensors to infrastructure IDs but I think this would only really matter if we have more that say 100 infrastructure instances. With less than 10 infrastructure instances I do not think this is a significant issue. Creating an additional map requires some small changes to how we instantiate each infrastructure instance and I think currently with less than 5 instances planned for concurrent deployment, I do not think it is a large issue. I think we can add a TODO statement to resolve this for larger deployments

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
paulbourelly999: No each send message sends to a different port.

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/DetectedObject.java
```diff
@@ -0,0 +1,249 @@
...
+
+    private Double[] positionCovariance = new Double[9];
+
+    private Vector3d velocity;
+
+    private Double[] velocityCovariance = new Double[9];
+
+    private Vector3d angularVelocity;
+
+    private Double[] angularVelocityCovariance = new Double[9];
```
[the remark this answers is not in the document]
paulbourelly999: Please read me response above.

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/Detector.java
```diff
@@ -0,0 +1,104 @@
...
+
+import java.io.Serializable;
+
+import org.eclipse.mosaic.lib.geo.CartesianPoint;
+
+public class Detector implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    private String sensorId;
+    private DetectorType type;
```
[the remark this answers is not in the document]
paulbourelly999: In Java where you have to define getters and setters I do not think javadoc comments for member variables are as important. Especially when they are pretty self explanatory like sensor id or Detector Type. I think it is enough to leave comments in the getter/setter methods that modify/return the members

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java
```diff
@@ -16,17 +16,19 @@
...
+import org.eclipse.mosaic.lib.geo.CartesianPoint;
+import org.eclipse.mosaic.lib.objects.detector.Detector;
 
 /**
  * InfrastructureInstance class represents a physical instance of an
- * infrastructure node in the simulated environment.
- * It contains information about the infrastructure node such as its ID,
+ * infrastructure instance in the simulated environment.
+ * It contains information about the infrastructure instance such as its ID,
  * location, target address, and ports.
```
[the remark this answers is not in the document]
paulbourelly999: I can add to this constructor comment but it is not really in the scope of my PR. Also a lot of it is pretty self explanatory like ID or location or address.

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/DetectorType.java
paulbourelly999: I do not think that is necessary either. The goal here is not to allow any format of the label to be translated into the enumeration but rather to provide a method for the JSON adapter to read the enumeration from the json message.

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/Orientation.java
paulbourelly999: As mentioned before I do not think a Javadoc comment for each member is required in Java where getters/setters already have that information about the member .

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/Size.java
paulbourelly999: As mentioned before I do not think a Javadoc comment for each member is required in Java where getters/setters already have that information about the member .

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/gson/DetectorTypeAdapter.java
paulbourelly999: This naming is inline with other GSON serialize and refers to the annotation https://github.com/usdot-fhwa-stol/cdasim/blob/35c45a53d499c7ba8901177ffd73f5b6a0d32fab/co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/DetectorType.java#L24

co-simulation/lib/mosaic-objects/src/test/java/org/eclipse/mosaic/lib/objects/detector/DetectorTypeTest.java
paulbourelly999: We have unit tests that cover it by testing Detector JSON serialization/deserialization

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/DetectedObject.java
```diff
@@ -0,0 +1,249 @@
...
+
+    private String sensorId;
+
+    private String projString;
+
+    private String objectId;
+
+    private CartesianPoint position;
+
+    private Double[] positionCovariance = new Double[9];
```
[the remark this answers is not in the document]
dan-du-car: make sense. Maybe replace the number 9 with a constant with a more intuitive name as I see number 9 is used at multiple places?

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/DetectorType.java
dan-du-car: I am getting at is that the format including the case and spaces. Do we have a warning to tell why a label is not a valid DetectorType. EG: `SemanticLidar` is a valid DetectorType, but `semanticlidar` is not a valid DetectorType.

co-simulation/fed/mosaic-infrastructure/pom.xml
paulbourelly999: I agree that it would have been better to include mosiac as a maven dependency but completely disagree with your statement about package structure being confusing. I think doing this kind of package structure is standard practice for extending Java frameworks/libraries. The package structure gives insight not only to what functionality the code inside provides but whether it is part of the core framework or not and what organization contributed this extension.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java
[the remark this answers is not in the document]
paulbourelly999: https://github.com/usdot-fhwa-stol/cdasim/blob/522233f0ffe87617f00a686883dcf8cd153b7c2f/co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java#L48C1-L49C1
https://github.com/usdot-fhwa-stol/cdasim/blob/522233f0ffe87617f00a686883dcf8cd153b7c2f/co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java#L55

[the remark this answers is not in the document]
paulbourelly999: In the constructor the parameters are described as follows:
```
    /**
     * Constructor for InfrastructureInstance
     * 
     * @param infrastructureId          the ID of the infrastructure instance.
     * @param targetAddress             the target IP address of the infrastructure instance.
     * @param rxMessagePort             the receive V2X message port of the infrastructure instance.
     * @param timeSyncPort              the receive time synchronization message port of the infrastructure.
     *                         
     * @param simulatedInteractionPort  the receive simulated interaction message port of the infrastructure
     *                                  instance.
     * @param location                  the location of the infrastructure instance in the
     *                                  simulated environment
     */ 
I guess I could add that the location is in the simulation map frame and not a geodetic location, but it seems pretty clear since it says `simulated environment` and the location is a cartesian point and not a geodetic point.
```

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java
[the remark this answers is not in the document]
paulbourelly999: I understand, in my opinion when doing these reviews, we should judge the documentation based on whether it is reasonable to expect someone foreign to the code to be able to read it and understand it without requiring an unreasonable amount of investigation. With that thought in mind I think documenting your methods declarations is a must/especially any public methods that others may use. Additionally, a combination of variable naming and inline comments should be able to give meaning to your variables. I do not think it is important to add a comment to a location variable explicitly saying it is the location, but if you have 5 location variables i think it is important to document inline or constructor what the difference between the 5 location variables are. In that spirit I could add documentation to the port members but I think in this case they are named well and have constructor documentation that makes this not completely necessary.

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/DetectedObject.java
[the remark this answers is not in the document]
paulbourelly999: Will do. Also add a comment.

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/DetectionType.java
```diff
@@ -0,0 +1,58 @@
...
+    
+    /**
+     * Returns the enum mapped from an String label.
+     *
+     * @param label string.
+     * @return the enum mapped from String label.
+     */
+    public static DetectionType fromLabel(String label) {
+        for (DetectionType type: DetectionType.values()) {
+            if (type.label.equals(label)) {
```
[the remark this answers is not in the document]
paulbourelly999: I can change this if this is important.

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/DetectorType.java
paulbourelly999: I guess I can fix this.

co-simulation/fed/mosaic-infrastructure/pom.xml
paulbourelly999: @kjrush if you have input on whether this package structure is confusing and also whether I should avoid adding to eclipse packages like I have in this PR and instead put my code in a fhwa library like carma utils one please let me know

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java
dan-du-car: This new DatagramSocket is to create a new UDP  socket and bind to any available port on the host machine according to: https://docs.oracle.com/javase/8/docs/api/java/net/DatagramSocket.html.
However, it seems that we want to create a Client side UDP socket and connect to remote infrastructure instance with IP (targetAddress) and the ports specified (rxMessagePort, timeSyncPort,simulatedInteractionPort ). 
This socket here does not tell which port and IP to send each type of message (V2x, TimeSync, SimulatedInteraction)?

dan-du-car: The bind() gives the vibe that we are creating a UDP server socket and trying to bind it to a particular port. But this infrastructure instance in fact acts as a series of UDP clients and send packet to remote infrastructure. Maybe we can just call this `socket = new DatagramSocket` inside the infrastructure instance initialization.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java
```diff
@@ -35,40 +37,48 @@ public class InfrastructureInstance {
...
-     * @param timeSyncPort     the time synchronization port of the infrastructure
-     *                         node
-     * @param location         the location of the infrastructure node in the
-     *                         simulated environment
-     */
-    public InfrastructureInstance(String infrastructureId, InetAddress targetAddress,
-            int rxMessagePort, int timeSyncPort, CartesianPoint location) {
+     * @param infrastructureId          the ID of the infrastructure instance.
+     * @param targetAddress             the target IP address of the infrastructure instance.
+     * @param rxMessagePort             the receive V2X message port of the infrastructure instance.
```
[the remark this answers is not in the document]
paulbourelly999: This is a object referring to the infrastructure instance, not the Ambassador. In this case it is the port on which the infrastructure instance expects to receive V2X messages, Time Sync Message, etc. This comes from the InfrastructureRegistration message which infrastructure sends to the Ambassador and leads to the creation of this object

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java
paulbourelly999: This is not really part of the scope for this PR but I can look into a better naming convention for this method. It is a single client though, just different ports which can be assigned at the datagrampacket level not the socket level.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java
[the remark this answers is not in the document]
dan-du-car: I am not super familiar with the registration, and timeSync. But the simulatedInteractionPort is used to send the detectedObject. I see the Ambassador has receiveDetectedObjectInteraction which will receive the detected object and call infrastructureManager which on the other hand call infrastructure instance to send the detected object to remote host and simulatedInteractionPort .

I see the Registration Receiver and Time message receiver, so I am assuming the data flow was similar?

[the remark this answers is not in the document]
dan-du-car: - Updating timeSync upon process advance grant: https://github.com/usdot-fhwa-stol/cdasim/blob/a837403a2042ee7991f69643e780aaf8bf5473f7/co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureMessageAmbassador.java#L331 and send timeSync message to infrastructure by calling the send sendTimeSyncMsg() from the infrastructure instance.
- Upon receiving the V2xIntersection, this line https://github.com/usdot-fhwa-stol/cdasim/blob/a837403a2042ee7991f69643e780aaf8bf5473f7/co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureMessageAmbassador.java#L201C43-L201C57 send V2x message to infrastructure by calling the sendInteraction() function from the infrastructure instance. 

If those are correct, I still think the rxMessagePort and timeSyncPort are used to send data to remote infrastructure listen on targetAddress and corresponding port. The comment on each of these ports are telling the opposite.

co-simulation/fed/mosaic-infrastructure/pom.xml
kjrush: For context, the rationale behind cloning the repo rather than simply linking the maven artifacts was simply that we had full intent to be modifying the core simulator packages in order to add CARLA as a whole new class of simulator. We don't truly need every package but I feel like a patchwork solution is likely more confusing than just maintaining our own fork.

Overall though, I agree with your assessment on the group name. The group name is intended to indicate maintainer as you said, such that you could have multiple versions of a package maintained by different groups with possibly different implementations or features. That does make me wonder if truly the correct solution here is to actually migrate all the packages to gov.dot.fhwa.stol, since we're maintaining this fork, but I'm not sure that's a question to be resolved here.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java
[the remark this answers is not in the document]
paulbourelly999: What would you prefer the wording in the comment to be.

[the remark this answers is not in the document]
dan-du-car: simulatedInteractionPort: The port used to send simulated interaction message.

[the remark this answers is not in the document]
paulbourelly999: To clarify, Infrastructure Instance is meant to hold the registration information for each V2X-Hub/Infrastructure Instance. When we say rxMessagePort is the port on V2X-Hub(Infrastructure Instance) that is listening for V2X messages. The same is true for time sync. You are thinking about the infrastructure instance as if it is a custom datasocket but it is a model capturing the information/connection to a registered external infrastructure instance (V2X_Hub). I think in the context of a Mosiac Ambassador this makes sense. It also make sense in the context of the infrastructureInstanceManager. But if you have some better wording you would like to suggest I am open to it.

dan-du-car (approved):

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java
kjrush: Maybe `connect` is a better name or something along those lines? the primary reason - if I recall correctly - for having a separate method to init the DatagramSocket was to avoid a possible failure in the constructor (e.g. the ports are already in use) which is just a weird state to have. Not strongly attached to this implementation though.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstance.java
```diff
@@ -167,27 +234,40 @@ public void bind() throws IOException {
...
     }
 
     /**
-     * Sends time sync data to the Infrastructure Device communications interface configured at construction time.
+     * Sends time sync data to the Infrastrucutre Instance Time Sync port.
+     * 
+     * @param data The binary data to transmit
+     * @throws IOException If there is an issue with the underlying socket object or methods
+     */
+    public void sendTimeSyncMsg(byte[] data) throws IOException {
```
kjrush, line 247: Have we defined a class for the Time Sync message? I feel like it might be cleaner to have this (and the interaction one below) take in some type more meaningful than a byte[]. I think it make sense in the v2x message case because byte[] is actually the real type of a V2XMessage, but the time data and (I imagine) the interaction data are more higher-level than that.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
kjrush: Agree with dan here, but also agree that it's not a big deal for small numbers. We're on order of O(n^2), but for small n that's still fine. I like your suggestion of adding a comment, a TODO in the doc comment would be great so that callers know the performance limitations.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
```diff
@@ -135,24 +151,79 @@ public V2xMessageTransmission onV2XMessageTx(InetAddress sourceAddr, CarmaV2xMes
...
+        }
+    }
+
+    /**
+     * Helper method to serialize message into JSON and encode as bytes.
+     * 
+     * @param message java object containing message information
+     * @return bytes encoded from JSON string representation of object.
+     */
+    private byte[] encodeMsg(Object message) {
```
kjrush, line 203: To echo above, it's not a big deal, but I'm not a huge fan of how broad this function signature is. If I'm tinkering in this class in the future, this function signature doesn't do much to inform me what type of stuff we're usually encoding or need to encode. I'd generally prefer narrow function signatures where possible.

co-simulation/fed/mosaic-infrastructure/src/main/java/org/eclipse/mosaic/fed/infrastructure/ambassador/InfrastructureInstanceManager.java
```diff
@@ -135,24 +151,79 @@ public V2xMessageTransmission onV2XMessageTx(InetAddress sourceAddr, CarmaV2xMes
...
+     *                from the ambassador side
+     * @throws IOException
+     */
+    public void onTimeStepUpdate(InfrastructureTimeMessage message) throws IOException {
+        if (managedInstances.size() == 0) {
+            log.debug("There are no registered instances");
+        }
+
+        for (InfrastructureInstance currentInstance : managedInstances.values()) {
+            currentInstance.sendTimeSyncMsg(encodeMsg(message));
```
kjrush, line 223: This is somewhere I think we could just make `sendTimeSyncMsg` consume InfrastructureTimeMessage like I noted prior, then the entire encodeMsg function is no longer needed (or at least not needed here) and then I think the API is a little clearer in terms of intended use.

kjrush (dismissed):
Few minor things, but overall looks good to me!

kjrush (approved):

PR #165, opened 2023-10-23
Update CARLA Ambassador to use XMLRPC Client

paulbourelly999:
# PR Details
## Description

Add logic to CARLA Ambassador to use XMLRPC Client to dynamically create sensor and retrieve detections. More specifically: the CARLA Ambassador will now create a CARLA XML RPC Client on startup. Its configuration parameters (constructor arguments) will be loaded from the carla_config.json configuration file for the CARLA Ambassador. This is currently limited to **carlaCDASimAdapterUrl** which tells the XMLRPC client where the XMLRPC server is hosted.
Next the CARLA Ambassador now listens for **Detector Registration Interactions** which are triggered by the Infrastructure Ambassador in response to infrastructure registration attempts that include sensors. When receiving a **Detector Registration Interaction**, the CARLA Ambassador will tell the CARLA XML RPC Client to make a `create_sensor` request for each detector registration interaction to create the sensors in CARLA. On each timestep, which should be at 100 ms intervals, the CARLA Ambassador will now also call `get_detections` for every successful `create_sensor` attempt. This means that all the created sensor in CARLA will have their detections polled at the same interval as the CARLA Ambassador runs `processTimeAdvanceGrant`. The detections retrieved from the CARLA CDA Sim Adapter via the XMLRPC client will be serialized into `DetectedObject`(s) from json and triggered as interactions for other listening Ambassadors, mainly the infrastructure ambassador.
## Related Jira Key
[CDAR-209
](https://usdot-carma.atlassian.net/browse/CDAR-209)
## Motivation and Context
Add CARLA Sensor pipeline for creating sensors and getting detections for CARMA-Streets

## How Has This Been Tested?
Unit testing 

## Types of changes

## Checklist:

- [x] All new and existing tests passed

### DIFF
--- a/co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
+++ b/co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
@@ -519,7 +519,10 @@ public class CarlaAmbassador extends AbstractFederateAmbassador {
         }
     }
 
-
+    /**
+     * Method to call XMLRPC method to create sensor on reception of DetectionRegistration interactions. 
+     * @param interaction Interaction triggered by Ambassadors attempting to create sensors in CARLA.
+     */
     private void receiveInteraction(DetectorRegistration interaction) {
         try {
             carlaXmlRpcClient.createSensor(interaction);
--- a/co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
+++ b/co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
@@ -218,6 +218,7 @@ public class CarlaAmbassador extends AbstractFederateAmbassador {
         try{
             URL xmlRpcServerUrl = new URL(carlaConfig.carlaCDASimAdapterUrl);
             carlaXmlRpcClient = new CarlaXmlRpcClient(xmlRpcServerUrl);
+            carlaXmlRpcClient.initialize();
         }
         catch (MalformedURLException m) 
         {
--- a/co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
+++ b/co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
@@ -362,7 +362,7 @@ public class CarlaAmbassador extends AbstractFederateAmbassador {
             // CARLA but it seems to increment every time processTimeAdvanceGrant is called
             rti.requestAdvanceTime(nextTimeStep + this.executedTimes, 0, (byte) 2);
             this.executedTimes++;
-            List<DetectedObjectInteraction> detectedObjectInteractions = new ArrayList<DetectedObjectInteraction>();
+            List<DetectedObjectInteraction> detectedObjectInteractions = new ArrayList<>();
             // Get all detections from all currently registered detectors.
             for (DetectorRegistration registration: registeredDetectors ) {
                 DetectedObject[] detections = carlaXmlRpcClient.getDetectedObjects( registration.getInfrastructureId() , registration.getDetector().getSensorId());
@@ -623,7 +623,7 @@ public class CarlaAmbassador extends AbstractFederateAmbassador {
                 }
             }
         } catch (Exception e) {
-            log.error("error occurs during sending messages to bridge: " + e.getMessage());
+            log.error("error occurs during sending messages to bridge: {}", e.getMessage());
         }
     }
 
@@ -649,7 +649,7 @@ public class CarlaAmbassador extends AbstractFederateAmbassador {
                 carlaConnection.getDataOutputStream().write(new byte[] { 0x07, 0x2f, 0x00, 0x00, 0x00, 0x00, 0x00 });
             }
         } catch (Exception e) {
-            log.error("error occurs during process received messages: " + e.getMessage());
+            log.error("error occurs during process received messages: {}",  e.getMessage());
         }
         return message.split(";");
     }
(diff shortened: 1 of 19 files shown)


co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/carlaconnect/CarlaXmlRpcClient.java
```diff
@@ -63,33 +71,35 @@ public void initialize(URL xmlRpcServerUrl)
...
+        else {
+            log.warn("XMLRpcClient is not connected to CARLA Adapter!");
+        }
     }
+
+    public DetectedObject[] getDetectedObjects(String infrastructureId ,String sensorId) throws XmlRpcException{
+        if (isConnected) {
+            Object[] params = new Object[]{infrastructureId, sensorId};
+            Object result = client.execute(GET_DETECTED_OBJECTS, params);
+            log.info("Detections from infrastructure {} sensor {} : {}", infrastructureId, sensorId, result);
```
MishkaMN, line 92: I am thinking this may get really big log if we detect the log the result continuously? Maybe we should move it into debug?

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/DetectedObject.java
```diff
@@ -37,15 +37,15 @@ public final class DetectedObject implements Serializable {
 
     private CartesianPoint position;
 
-    private Double[] positionCovariance = new Double[9];
+    private Double[][] positionCovariance = new Double[3][3];
```
MishkaMN, line 40: Just curious whether if there is any change in rest of the pipeline until this message reaches SDSM convertor in carma-street

MishkaMN (changes requested):

co-simulation/lib/mosaic-objects/src/main/java/org/eclipse/mosaic/lib/objects/detector/DetectedObject.java
paulbourelly999: I do not believe so. V2X-Hub treats the JSON as a string and passes it on and the SensorDetectedObject has not been implemented on CARMA-Streets yet.

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/carlaconnect/CarlaXmlRpcClient.java
paulbourelly999: I can move this to debug.

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/carlaconnect/CarlaXmlRpcClient.java
```diff
@@ -30,7 +30,8 @@
 import com.google.gson.Gson;
 
 /**
- * This is a class uses xmlrpc to connect with CARLA CDASim adapter service
+ * This is a class uses xmlrpc to connect with CARLA CDASim Adapter. It includes calls
```
MishkaMN, line 33: class that* uses

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/carlaconnect/CarlaXmlRpcClient.java
```diff
@@ -30,7 +30,8 @@
 import com.google.gson.Gson;
 
 /**
- * This is a class uses xmlrpc to connect with CARLA CDASim adapter service
+ * This is a class uses xmlrpc to connect with CARLA CDASim Adapter. It includes calls 
+ * to dynamically create senors and get detected objects from created sensors.
```
MishkaMN, line 34: sensors*

MishkaMN (approved):

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
```diff
@@ -351,15 +357,29 @@ public synchronized void processTimeAdvanceGrant(long time) throws InternalFeder
                 nextTimeStep += carlaConfig.updateInterval * TIME.MILLI_SECOND;
                 isSimulationStep = false;
             }
+            // TODO: What is this. Why are we request a time advance based on this counter and 
+            // what is it counting. It is labeled as counting the times we attempt to connect to 
+            // CARLA but it seems to increment every time processTimeAdvanceGrant is called
             rti.requestAdvanceTime(nextTimeStep + this.executedTimes, 0, (byte) 2);
```
paulbourelly999, line 363: @kjrush Do you have any information why were are requesting time advance like this with a continuously incremented counter? If you have any information about executedTimes and why its used in our requestAdvanceTime.

kjrush: I did a quick skim of the rest of this PR, but so far focused on this.

I don't really see what purpose this executed times has, to be honest. It might just be entirely unnecessary. For the record, it doesn't increment infinitely. It's actually reset later on in the code inside the `receiveInteraction(SimulationStepResponse)` method. The interactions being received there seem to be generated whenever SUMO provides a new TraCI update, so it should be frequently, so this should never go above a small value, maybe 10 at most.

But even 10 doesn't make much sense, considering this `requestAdvanceTime` arg is in milliseconds, and there's no multiplicative factor applied to this `executedTimes` value (unlike `carlaConfig.updateInterval` above), so it's small enough it probably doesn't actually do anything. I imagine it was implemented like this with the goal that if we don't have any fresh data from sumo, we should wait longer to delay the CARLA updates until we do have fresh data. But I'm not sure that's needed and it seems unlikely this is actually accomplishing that.

I think you were right to be questioning about this line, but in the end I'm not sure it matters one way or the other. I'd be in favor of just removing it, but keeping this line in our mental bookmarks in case we run into issues with the ambassador. If we hit those issues we'll have this top of mind to re-investigate.

MishkaMN (approved):

PR #177, opened 2023-11-10
Change location of scenarios that will be deployed in docker image.

### LINKED ISSUE #175: co-simulation scenarios need to be manually extracted in the docker build
### Summary

The co-simulation scenarios are currently being unzipped in the /opt/carma-simulation project directory.
These resources should be made available to the project through the maven build.

### Commit

Develop

### Expected Behavior

Scenario files can be kept in the default mosaic folder and unpacked from there after the build.

### Actual Behavior

The scenario's need to be extracted into the built project.

### Steps to Reproduce the Actual Behavior

1. Comment out this manual extraction into /opt/carma https://github.com/usdot-fhwa-stol/cdasim/blob/36ec8872d2b0eaa1cb878a793c23b7b6998d48ff/docker/install.sh#L109
2. Rebuild the cdasim image
3. Check if resources are available within /opt/carma-simulation

### Related Work

_No response_

kjrush:
Scenarios will now be deployed from co-simulation/bundle/src/assembly/scenarios rather than from sample_scenarios.zip

# PR Details
## Description

Changes the docker build process to automatically deploy scenario files found in src/bundle/assembly/resources/scenarios instead of relying on a pre-zipped archive of scenario files.

## Related GitHub Issue

Closes #175

## Motivation and Context

This makes the behavior for adding a scenario more intuitive and lets us maintain scenarios as plain files instead of a zip archive.

## How Has This Been Tested?

Ran locally after building fresh docker image and confirmed newly created test scenario files were deployed at /opt/carma-simulation/scenarios in the final image.

## Types of changes

## Checklist:

- [x] I have added any new packages to the sonar-scanner.properties file

### DIFF
--- a/docker/install.sh
+++ b/docker/install.sh
@@ -106,7 +106,7 @@ cp bundle-22.1-SNAPSHOT.jar /opt/carma-simulation
 
 # Deploy scenario files
 cd /home/carma/src/co-simulation
-unzip sample_scenario.zip -d /opt/carma-simulation/scenarios
+unzip bundle/target/bundle-*.jar 'scenarios/*' -d /opt/carma-simulation/scenarios
 
 # Install NS-3
 cd "/opt/carma-simulation/bin/fed/ns3/"


docker/install.sh
```diff
@@ -106,7 +106,7 @@ cp bundle-22.1-SNAPSHOT.jar /opt/carma-simulation
 
 # Deploy scenario files
 cd /home/carma/src/co-simulation
-unzip sample_scenario.zip -d /opt/carma-simulation/scenarios
+unzip bundle/target/bundle-*.jar 'scenarios/*' -d /opt/carma-simulation/scenarios
```
adev4a, line 109: Looks like copies both the unzipped scenarios folder and its files to the location given. To replace it we need to update the destination path to /opt/carma-simulation/
With this we'll also need to supply an overwrite argument to unzip

adev4a: Thinking about this a little more, can we not use maven to copy the resources to that directory while building? 
There is this output resources tag which can be used to do that https://maven.apache.org/plugins/maven-resources-plugin/copy-resources-mojo.html#outputdirectory

paulbourelly999:
Should this change impact VRU at all?

PR #200, opened 2024-02-16
CARMA Cloud ambassador integration

kruegersp:
# PR Details
## Description

This PR primarily focuses on merging a branch into develop, which encompasses the integration testing of CARMA Cloud with CDASim. The purpose of this integration test was to assess the communication between CARMA Cloud and CDASim, ensuring that:

All components operate correctly within CDASim without encountering any significant problems.
CARMA Cloud successfully registers with Ambassador.
CARMA Cloud receives and processes time synchronization messages.

## Related Jira Key

[CXC-10](https://usdot-carma.atlassian.net/jira/software/c/projects/CXC/issues/CXC-10)

## Types of changes

## Checklist:

[CXC-10]: https://usdot-carma.atlassian.net/browse/CXC-10?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


paulbourelly999:
Please break this PR into smaller independent pieces of functionality like : 
- CARMA Cloud Ambassador
- Sumo scenario configuration
- CARLA Scenario configuration

Also lets not create a 3rd Town04 CDASim scenario that is just a copy of the others. It would require unnecessary maintenance of three separate scenarios.

co-simulation/bundle/src/assembly/resources/etc/runtime.json
```diff
@@ -121,6 +121,21 @@
...
+            "classname": "org.eclipse.mosaic.fed.carmacloud.ambassador.CarmaCloudMessageAmbassador",
+            "configuration": "carma-cloud_config.json",
+            "priority": 50,
+            "host": "local",
+            "port": 0,
+            "deploy": false,
+            "start": false,
+            "subscriptions": [
+                "V2xMessageReception",
+                "DetectedObjectInteraction"
```
paulbourelly999, line 135: Does CarmaCloud care about either of these interactions. What interactions does CARMA Cloud handle.

paulbourelly999:
@kruegersp Please address code smells, duplication and code coverage from sonar scan report.

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudInstance.java
```diff
@@ -0,0 +1,93 @@
...
+	{
+		HttpURLConnection oHttp = (HttpURLConnection)new URL(m_sCarmaCloudUrl).openConnection();
+		oHttp.setRequestMethod("POST");
+		oHttp.setDoOutput(true);
+		try (DataOutputStream oOut = new DataOutputStream(oHttp.getOutputStream()))
+		{
+			oOut.write(new Gson().toJson(oMsg).getBytes());
+		}
+		if (oHttp.getResponseCode() != 200)
+			throw new IOException();
```
paulbourelly999, line 91: This exception is misleading and contains no message

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudInstanceManager.java
```diff
@@ -0,0 +1,108 @@
...
+	 * 
+	 */
+	public void onNewRegistration(CarmaCloudRegistrationMessage registration)
+	{
+		if (!managedInstances.containsKey(registration.getId()))
+		{
+			CarmaCloudInstance tmp = new CarmaCloudInstance(registration.getId(), registration.getUrl());
+			managedInstances.put(registration.getId(), tmp);
+		}
+		else
```
paulbourelly999, line 59: Please always include `{}`  since it is easy to miss if they are not included.

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudMessageAmbassador.java
```diff
@@ -0,0 +1,224 @@
...
+	 */
+	@Override
+	public synchronized void processTimeAdvanceGrant(long time)
+		throws InternalFederateException
+	{
+		// Process the time advance only if the time is equal or greater than the next
+		// simulation time step
+		log.info("processTimeAdvanceGrant {} {}", currentSimulationTime, time);
+		if (time < currentSimulationTime)
+				return;
```
paulbourelly999, line 150: Please add brackets to this if statement

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudMessageAmbassador.java
```diff
@@ -0,0 +1,224 @@
...
+			// Advance the simulation time
+			currentSimulationTime += carmaCloudConfiguration.updateInterval * TIME.MILLI_SECOND;
+
+			// Request the next time advance from the RTI
+			log.info("Requesting timestep updated to  {}.", currentSimulationTime);
+			rti.requestAdvanceTime(currentSimulationTime, 0, (byte) 2);
+		}
+		catch (IllegalValueException e)
+		{
+				throw new InternalFederateException(e);
```
paulbourelly999, line 178: This indentation seems off.

co-simulation/fed/mosaic-carma-cloud/src/test/java/org/eclipse/mosaic/fed/carmacloud/CarmaCloudRegistrationReceiverTest.java
```diff
@@ -0,0 +1,81 @@
...
+ 
+         // Send the test message to the receiver
+         InetAddress address = InetAddress.getLocalHost();
+         sendSocket.connect(new InetSocketAddress(address, TEST_PORT), 10000);
+         DataOutputStream out = new DataOutputStream(sendSocket.getOutputStream());
+         out.writeUTF(json);
+         out.close();
+ 
+         // Wait for the message to be received
+         Thread.sleep(1000);
```
paulbourelly999, line 70: Using thread sleep in a unit test is bad practice. It makes the test take much longer than necessary and is hardware specific. Consider if there is an event driven way to avoid this. If absolutely necessary here is a post explaining the downfalls of Thread sleep in a unit test and a java library to use  instead https://medium.com/cloud-native-daily/why-is-testing-with-thread-sleep-a-bad-idea-and-what-can-you-do-instead-3aad1d26a787

co-simulation/fed/mosaic-carma-cloud/src/test/java/org/eclipse/mosaic/fed/carmacloud/CarmaCloudRegistrationReceiverTest.java
```diff
@@ -0,0 +1,81 @@
...
+ import java.net.InetAddress;
+ import java.net.InetSocketAddress;
+ import java.net.Socket;
+ import java.util.List;
+ 
+ import org.junit.After;
+ import org.junit.Before;
+ import org.junit.Test;
+ 
+ public class CarmaCloudRegistrationReceiverTest {
```
paulbourelly999, line 31: This test is not really a unit test. You are actually creating a real socket and sending information to it. Consider using Mocks instead

docker/install_dependencies.sh
```diff
@@ -61,7 +61,8 @@ CARLA_TAR="CARLA_0.9.10.tar.gz"
 cd /home/carma/src/
 if [[ ! -f "$CARLA_TAR" ]]; then
     echo "!!! $CARLA_TAR not present in the installation directory, downloading automatically instead. This could take a long time, consider downloading the file manually and placing it in the installation directory. !!!"
-    wget -q "https://carla-releases.s3.eu-west-3.amazonaws.com/Linux/CARLA_0.9.10.tar.gz"
+    wget -q "https://carla-releases.s3.us-east-005.backblazeb2.com/Linux/CARLA_0.9.10.tar.gz"
```
paulbourelly999, line 64: This change already exists in develop. Not sure why it also exists in this PR.

co-simulation/fed/mosaic-carma-cloud/pom.xml
```diff
@@ -0,0 +1,105 @@
...
+          <version>22.1-SNAPSHOT</version>
+          <scope>compile</scope>
+      </dependency>
+      <dependency>
+          <groupId>jakarta.xml.bind</groupId>
+          <artifactId>jakarta.xml.bind-api</artifactId>
+          <version>2.3.2</version>
+      </dependency>
+      <dependency>
+          <groupId>org.glassfish.jaxb</groupId>
```
paulbourelly999, line 81: What is this dependency necessary for?

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudInstanceManager.java
```diff
@@ -0,0 +1,108 @@
...
+	 */
+	public void onTimeStepUpdate(TimeSyncMessage message)
+		throws IOException
+	{
+		if (managedInstances.isEmpty())
+			log.debug("There are no registered instances");
+		else
+		{
+			for (CarmaCloudInstance currentInstance : managedInstances.values())
+				currentInstance.sendTimeSyncMsg(message);
```
paulbourelly999, line 80: Please always include curly braces for if, while and for  statements. Leaving them out is error prone and difficult to read.

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudInstanceManager.java
```diff
@@ -0,0 +1,108 @@
...
+	/**
+	 * External helper function to allow the ambassador to check if a given vehicle
+	 * ID is a registered CARMA Platform instance
+	 * 
+	 * @param carmaCloudId The id to check
+	 * @return True if managed by this object (e.g., is a registered CARMA Platform
+	 *         vehicle). false o.w.
+	 */
+	public boolean checkIfRegistered(String carmaCloudId) {
+			return managedInstances.keySet().contains(carmaCloudId);
```
paulbourelly999, line 94: This applies to multiple files but it seems the indentation is inconsistent even in a single file. Please make sure you format the files.

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudInstanceManager.java
```diff
@@ -0,0 +1,108 @@
...
+
+	/**
+	 * External helper function to allow the ambassador to check if a given vehicle
+	 * ID is a registered CARMA Platform instance
+	 * 
+	 * @param carmaCloudId The id to check
+	 * @return True if managed by this object (e.g., is a registered CARMA Platform
+	 *         vehicle). false o.w.
+	 */
+	public boolean checkIfRegistered(String carmaCloudId) {
```
paulbourelly999, line 93: Does anything call this method

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudMessageAmbassador.java
```diff
@@ -0,0 +1,224 @@
...
+	 * This method is called by the AbstractFederateAmbassador when the RTI grants a
+	 * time advance to the federate.Any unprocessed interactions are forwarded to
+ the federate using the processInteraction method before this call is made.
+	 *
+	 * @param time The timestamp (in nanoseconds) indicating the time to which the federate can
+	 *             advance its local time.
+	 * @throws InternalFederateException
+	 */
+	@Override
+	public synchronized void processTimeAdvanceGrant(long time)
```
paulbourelly999, line 143: I know we are not great about this currently in our ambassadors but we should be setting our current simulation time to the time provided here. Otherwise by incrementing incorrectly we could introduce an offset.

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudMessageAmbassador.java
```diff
@@ -0,0 +1,224 @@
...
+
+			// Request the next time advance from the RTI
+			log.info("Requesting timestep updated to  {}.", currentSimulationTime);
+			rti.requestAdvanceTime(currentSimulationTime, 0, (byte) 2);
+		}
+		catch (IllegalValueException e)
+		{
+				throw new InternalFederateException(e);
+		}
+		catch (IOException e1)
```
paulbourelly999, line 180: Nit: Since the previous catch statement and this one do not share scope we can call both exceptions `e`.  Not high priority but adding weird unnecessary naming conventions like this makes for harder to read code.

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudRegistrationMessage.java
```diff
@@ -0,0 +1,71 @@
...
+
+
+	/**
+	 * Returns the ID of the CARMA Cloud instance
+	 * 
+	 * @return String the ID of the CARMA Cloud instance
+	 */
+	public String getId()
+	{
+			return id;
```
paulbourelly999, line 62: Nit: indentation

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudInstanceManager.java
```diff
@@ -0,0 +1,108 @@
...
+		{
+			for (CarmaCloudInstance currentInstance : managedInstances.values())
+				currentInstance.sendTimeSyncMsg(message);
+		}
+	}
+
+
+	/**
+	 * External helper function to allow the ambassador to check if a given vehicle
+	 * ID is a registered CARMA Platform instance
```
paulbourelly999, line 87: Wrong comment

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudRegistrationReceiver.java
```diff
@@ -0,0 +1,139 @@
...
+	public void stop()
+	{
+		running.set(false);
+		if (m_oSrvr != null)
+		{
+			try
+			{
+				m_oSrvr.close();
+			}
+			catch (Exception oEx)
```
paulbourelly999, line 114: Nit: Why do we have some many different local variable names for the same thing.

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudRegistrationReceiver.java
```diff
@@ -0,0 +1,139 @@
...
+	 * 
+	 * @return The list of received outbound message from all Infrastructure Device
+	 *         instances since last call of this method
+	 */
+	public List<CarmaCloudRegistrationMessage> getReceivedMessages()
+	{
+		List<CarmaCloudRegistrationMessage> output = new ArrayList<>();
+		synchronized (rxQueue)
+		{
+			output.addAll(rxQueue);
```
paulbourelly999, line 134: Nit: I dont think the add all operation needs to be inside the synchronized statement. Only modifications to the rxQueue need to be inside this statement

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudMessageAmbassador.java
chengyuan0124: Do you mean we should do `currentSimulationTime = time` rather than `currentSimulationTime += carmaCloudConfiguration.updateInterval * TIME.MILLI_SECOND;`

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudRegistrationReceiver.java
chengyuan0124: I believe @kruegersp was directly copied from [CARMA Ambassador](https://github.com/usdot-fhwa-stol/cdasim/blob/b7c197777637e88f31e165329180218de543b877/co-simulation/fed/mosaic-carma/src/main/java/org/eclipse/mosaic/fed/carma/ambassador/CarmaRegistrationReceiver.java#L102-L103) which was implemented by Kyle. @kjrush do you think this modification will proceed without any issues?

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudInstance.java
kruegersp: Added status code and response message detail to IOException.

co-simulation/fed/mosaic-carma-cloud/src/test/java/org/eclipse/mosaic/fed/carmacloud/CarmaCloudRegistrationReceiverTest.java
kruegersp: Removed. CARMA Cloud registration uses synchronous TCP, not asynchronous UDP like other modules, so sleep is not necessary.

co-simulation/fed/mosaic-carma-cloud/pom.xml
kruegersp: It appears Cheng fixed this.

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudRegistrationReceiver.java
kruegersp: Are we certain the synchronization isn't to prevent rxQueue modification while the addAll operation iterates?

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudInstanceManager.java
paulbourelly999: Does any non-test code call this method.

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudRegistrationReceiver.java
```diff
@@ -0,0 +1,139 @@
...
+        		DataInputStream oIn = new DataInputStream(oSock.getInputStream());
+
+				// parse message
+				CarmaCloudRegistrationMessage parsedMessage = 
+					new Gson().fromJson(oIn.readUTF(), CarmaCloudRegistrationMessage.class);
+
+				// Enqueue message for processing on main thread
+				synchronized (rxQueue)
+				{
+					log.info("New CARMA Cloud instance '{}' received with CARMA Cloud Registration Receiver.", parsedMessage.getId());
```
paulbourelly999, line 90: This log statement does not need to be inside the synchronized block right? I think it is best practice to minimize the amount of code inside the synchronized block to limit how much time a given resource is locked.

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudRegistrationReceiver.java
paulbourelly999: I am referring to the the name of the exception variable. Why not just always use e or ex? It seems every catch statement calls the exception something different. It is kind of nit picky so address at your discretetion but it is kind of the equivalent of like 
```java
for (int i = 0; i < 5; i++) {
  System.out.println(i);
}
...
for (int count = 0; i < 5; i++) {
  System.out.println(count);
}
...
for (int index = 0; i < 5; i++) {
  System.out.println(index);
}
```

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudRegistrationReceiver.java
paulbourelly999: Hmm good point bryan. That is probably correct

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudRegistrationReceiver.java
kruegersp: Moved after synchronization block.

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudInstanceManager.java
kruegersp: Nothing in the CARMA Cloud ambassador code calls it. It was copied from the Infrastructure ambassador code. Should we remove it?

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudRegistrationReceiver.java
kruegersp: Fixed.

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudInstanceManager.java
kruegersp: Removed.

co-simulation/fed/mosaic-carma-cloud/src/test/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudMessageAmbassadorTest.java
```diff
@@ -0,0 +1,133 @@
...
+    public void testInitialize() throws InternalFederateException, IllegalValueException {
+        // Test initialize method
+        ambassador.initialize(0, 100 * TIME.SECOND);
+        verify(rtiMock, times(1)).requestAdvanceTime(0L, 0L, ((byte) 1));
+        // cleanup threads and sockets
+        ambassador.close();
+    }
+
+//    @Test
+//    public void testProcessInteraction() throws InternalFederateException {
```
paulbourelly999, line 117: If this test is not being used, lets remove it. If it is being used lets fix and uncomment it

co-simulation/fed/mosaic-carma-cloud/src/test/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudRegistrationReceiverTest.java
```diff
@@ -0,0 +1,93 @@
...
+ 
+ public class CarmaCloudRegistrationReceiverTest {
+ 
+     private static final int TEST_PORT = 1617;
+ 
+//     private Socket sendSocket;
+//     private CarmaCloudRegistrationReceiver receiver;
+ 
+//     @Before
+//     public void setup() throws Exception {
```
paulbourelly999, line 45: As mentioned above if these methods are no longer needed, lets remove them. If they are needed lets fix them and uncomment them.

co-simulation/bundle/src/assembly/resources/etc/runtime.json
```diff
@@ -121,6 +121,20 @@
...
+            "id": "carma-cloud",
+            "classname": "org.eclipse.mosaic.fed.carmacloud.ambassador.CarmaCloudMessageAmbassador",
+            "configuration": "carma-cloud_config.json",
+            "priority": 50,
+            "host": "local",
+            "port": 0,
+            "deploy": false,
+            "start": false,
+            "subscriptions": [
+                "V2xMessageReception"
```
paulbourelly999, line 134: @chengyuan0124 Does the carma-cloud ambassador do anything with V2XMessageReception interactions?

co-simulation/fed/mosaic-carma-cloud/pom.xml
paulbourelly999: What has been fixed. I was asking what this dependency is used for. Is it being used somewhere

co-simulation/bundle/src/assembly/resources/etc/runtime.json
chengyuan0124: It doesn't, but the rule for subscriptions seems to require at least one entry in this section. I'll double-check to see if we can use an empty string here; otherwise, we might need to leave it as is.

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudMessageAmbassador.java
```diff
@@ -0,0 +1,228 @@
...
+			// nanoseconds to milliseconds for CarmaCloudTimeMessage
+			TimeSyncMessage timeSyncMessage = new TimeSyncMessage(currentSimulationTime / 1000000L, timeSyncSeq);
+			carmaCloudInstanceManager.onTimeStepUpdate(timeSyncMessage);
+
+			// Advance the simulation time
+			currentSimulationTime = time;
+
+			// Request the next time advance from the RTI
+			log.info("Requesting timestep updated to  {}.", currentSimulationTime + carmaCloudConfiguration.updateInterval* TIME.MILLI_SECOND);
+			rti.requestAdvanceTime(currentSimulationTime, 0, (byte) 2);
```
paulbourelly999, line 178: This is incorrect. we need to be requesting currentSimulationTime + step size

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudRegistrationReceiver.java
```diff
@@ -0,0 +1,139 @@
...
+ * Worker thread Runnable for operating a listen socket to receive outbound
+ * CARMA Cloud Messages from CARMA Cloud instances
+ * This {@link Runnable} instance will operate a UDP socket to subscribe to
+ * packets from a CARMA Cloud adapter. 
+ * Upon receiving a packet, it will be queued for the primary thread
+ * to process the data once it ticks to a simulation processing step
+ */
+public class CarmaCloudRegistrationReceiver implements Runnable
+{
+	private static final int LISTEN_PORT = 1617; // which port for CARMA Cloud?
```
paulbourelly999, line 43: Is this comment still relevant?

co-simulation/fed/mosaic-carma-cloud/src/main/java/org/eclipse/mosaic/fed/carmacloud/configuration/CarmaCloudConfiguration.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * The CARMA Cloud Message Ambassador configuration class.
+ */
+public class CarmaCloudConfiguration implements Serializable
+{
+	private static final long serialVersionUID = 1705520136000000000L;
+
+
+	@JsonAdapter(TimeFieldAdapter.LegacyMilliSeconds.class)
+	public Long updateInterval = 100L;
```
paulbourelly999, line 30: Should the CarmaCloud Ambassador include configuration for IP and Port of CARMA Cloud instance?

co-simulation/fed/mosaic-carma-cloud/src/test/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudInstanceTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+import gov.dot.fhwa.saxton.TimeSyncMessage;
+
+public class CarmaCloudInstanceTest {
+
+    private CarmaCloudInstance instance;
+    private StringBuilder messageBuf = new StringBuilder();
+    private HttpServer oSrvr;
+
+    class TimeSyncHandler implements HttpHandler {
```
paulbourelly999, line 44: Nit: We should probably leave a comment for the unit test to remove the HTTP Server and use mocks instead.

paulbourelly999 (changes requested):
Left some small comments.

co-simulation/fed/mosaic-carma-cloud/src/test/java/org/eclipse/mosaic/fed/carmacloud/ambassador/CarmaCloudRegistrationReceiverTest.java
paulbourelly999: This is still an issue.

paulbourelly999 (changes requested):
Please remove comment/unused unit tests. Then I am good with this PR.

paulbourelly999 (approved):

PR #205, opened 2024-03-19
Feature/add time sync logs

MishkaMN:
# PR Details
## Description

Adds logs to support https://github.com/usdot-fhwa-stol/carma-analytics-fotda/pull/43
## Related Jira Key
CDAR-774

## Motivation and Context
Data Analysis

## Types of changes

## Checklist:

### DIFF
--- a/co-simulation/fed/mosaic-sumo/src/main/java/org/eclipse/mosaic/fed/sumo/ambassador/AbstractSumoAmbassador.java
+++ b/co-simulation/fed/mosaic-sumo/src/main/java/org/eclipse/mosaic/fed/sumo/ambassador/AbstractSumoAmbassador.java
@@ -1202,14 +1202,11 @@ public abstract class AbstractSumoAmbassador extends AbstractFederateAmbassador 
         // A script to validate time synchronization of tools in CDASim currently relies on the following
         // log line. TODO: This line is meant to be removed in the future upon completion of this work:
         // https://github.com/usdot-fhwa-stol/carma-analytics-fotda/pull/43
-        if (log.isDebugEnabled())
+        if (log.isDebugEnabled() && (!receivedSimulationStep && firstTimePrintingTime))
         {
-            if (!receivedSimulationStep && firstTimePrintingTime)
-            {
-                long millis = System.currentTimeMillis();
-                log.info("Simulation Time: {} where current system time is: {} and nextTimeStep: {} and ambasador id: {}", (int) (time/1e6), millis, nextTimeStep, getId());
-                firstTimePrintingTime = false;
-            }
+            long millis = System.currentTimeMillis();
+            log.info("Simulation Time: {} where current system time is: {} and nextTimeStep: {} and ambasador id: {}", (int) (time/1e6), millis, nextTimeStep, getId());
+            firstTimePrintingTime = false;
         }
 
         if (time > lastAdvanceTime) {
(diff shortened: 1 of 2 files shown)


co-simulation/fed/mosaic-sumo/src/main/java/org/eclipse/mosaic/fed/sumo/ambassador/AbstractSumoAmbassador.java
```diff
@@ -1229,6 +1242,7 @@ public synchronized void processTimeAdvanceGrant(long time) throws InternalFeder
                 rti.triggerInteraction(simulationStepResult.getTrafficDetectorUpdates());
                 this.rti.triggerInteraction(simulationStepResult.getTrafficLightUpdates());
                 receivedSimulationStep = false;
+                firstTimePrintingTime = true;
```
paulbourelly999, line 1245: Why do you need both of these variables. `receivedSimulationStep` should be `!firstTimePrintingTime`

co-simulation/rti/mosaic-rti-core/src/main/java/org/eclipse/mosaic/rti/time/SequentialTimeManagement.java
```diff
@@ -87,6 +93,18 @@ public void runSimulation() throws InternalFederateException, IllegalValueExcept
             if (ambassador != null) {
                 federation.getMonitor().onBeginActivity(event);
                 long startTime = System.currentTimeMillis();
+
+                if (!logging_map.containsKey(event.getFederateId())) // first time printing, then directly save the last simulation time requested and print
```
paulbourelly999, line 97: This should all be avoidable by setting log level to info. All this should be inside an if for log level debug or trace and we may want to leave a comment to consider reworking this.  What you really want is just data about when each federate progresses in time. This is probably not a long term solution.

paulbourelly999 (changes requested):
See comments.

co-simulation/rti/mosaic-rti-core/src/main/java/org/eclipse/mosaic/rti/time/SequentialTimeManagement.java
paulbourelly999: You could even just make this whole block a method that takes in event  and start time.

paulbourelly999:
Address code smells.

co-simulation/fed/mosaic-sumo/src/main/java/org/eclipse/mosaic/fed/sumo/ambassador/AbstractSumoAmbassador.java
MishkaMN: Because they are meant for different cases. 
`receivedSimulationStep` is flagged when `carla` returns with some handshake to sumoAmbassador. However, sumoAmbassador is called many times before that by MOSAIC. 
`firstTimePrintingTime` is meant to track the first time ever sumoAmbassador is called to progressTime. And that is by design. 
In other words, we can't reuse the same variable, because if we pick `!receivedSimulationStep`, the log will still end up printing many times in almost like 1ms apart. This `firstTimePrintingTime` is meant to avoid such excessive logging.

co-simulation/rti/mosaic-rti-core/src/main/java/org/eclipse/mosaic/rti/time/SequentialTimeManagement.java
MishkaMN: Addressed

paulbourelly999 (approved):

PR #233, opened 2024-12-30
TT 113-Create-SUMO-background-traffic

mpalash95:
# PR Details
## Description

Creating relevant scenario, routing, sumo files for background traffic for Town 05. 

Possible options:
1. https://sumo.dlr.de/docs/Tools/Trip.html  to generate random route files given input parameters
2. https://sumo.dlr.de/docs/od2trips.html  (origin destination) allows you to configure origin and destination as well as intermediate edges to use.
3. https://sumo.dlr.de/docs/Definition_of_Vehicles%2C_Vehicle_Types%2C_and_Routes.html#incomplete_routes_trips_and_flows similar to od2trips but allows for definition of traffic flows (multiple vehicles that follow an origin/destination route). 

In this PR:
1. Option 1 has been selected and  background traffic route file for Town05 has been generated  using sumo tools **randomTrips.py**.
2. A bash script genrate_random_traffic.sh has been written to generate background traffic for given sumo network of Town05 which can be utilized to generate background traffic for other maps and uses case's requirement
3. A new directory named tools has been created to store the script genrate_random_traffic.sh under Co-simulation Directory

## Related Jira Key

TT-113

## How Has This Been Tested?
1. The generated background route file has been tested by running a Sumo simulation with the route file
Simulation in Sumo with the background route file

## Types of changes

## Checklist:

### DIFF
--- a/co-simulation/tools/geneate_background_traffic.sh
+++ b/co-simulation/tools/geneate_background_traffic.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+#This script will generate random traffic to be used in Sumo-carla co simulation for a any Sumo Network File given the simulation duration, number of vehicles and a random number seed
+# This script requires  installation of Sumo tools randomTrips.py and Python
+echo "Sumo Network Files in this directoryi:"
+ls *.net.xml
+# Prompt for user inputs
+echo 
+echo "Enter the Sumo Network File Name:"
+read net_file
+
+echo "Output File Name:"
+read out_file
+
+echo "Enter Duration of the Simulation:"
+read sim_duration
+
+echo "How many vehicles you want to instantiate?:"
+read veh_number
+
+echo "Enter Random Number Seed (Different Seed will genrate different traffic):"
+read seed
+
+# Calculate the rate of vehicle generation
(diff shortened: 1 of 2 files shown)


co-simulation/tools/geneate_background_traffic.sh
```diff
@@ -0,0 +1,69 @@
...
+read sim_duration
+
+echo "How many vehicles you want to instantiate?:"
+read veh_number
+
+echo "Enter Random Number Seed (Different Seed will genrate different traffic):"
+read seed
+
+# Calculate the rate of vehicle generation
+rate=$(echo "$sim_duration / $veh_number" | bc -l)
```
kjrush, line 24: Minor thought, but might be worth making these command-line arguments, and then taking interactive can be a fallback if not all the arguments are passed in. Doesn't have to be fancy even, could just assume that the args are always passed in the same order. And then if the count of args doesn't match expected, you run this interactive logic.

Only mentioned because this might conceivably be integrated eventually as part of a larger tool or automated pipeline and passing commandline arguments works way better for that than interactive input does.

mpalash95 (commented):
Changed the code to take command line arguments

DokurOmkar (approved):

PR #250, opened 2025-08-04
CDAS-65 Initial GUI for Automated Testing Tool

chengyuan0124:
# PR Details
## Description

This PR introduces a PySide6-based GUI for managing CDASim. The GUI allows users to browse a cdasim-config repository, select valid configurations, set up map and route files, pull Docker images, build and set configurations, and start/stop simulations. Key features include:

- Toolbar with buttons to browse the repo, open /opt/carma/logs, and open /opt/carma-simulation/logs.
- Dropdown to select valid configs, ensuring no default selection triggers setup prematurely.
- Horizontal alignment of Start/Stop buttons for better UX.
- Robust error handling for stop_simulation using multiprocessing to prevent segmentation faults.
- Automatic creation of /tmp/.env to resolve script errors.
- Logging to both GUI (with colored error/warning messages) and terminal for debugging.

## Types of changes

## Checklist:


GUI/src/gui_qt.py
```diff
@@ -0,0 +1,458 @@
...
+            while not self.log_queue.empty():
+                message, tag = self.log_queue.get_nowait()
+                self.log_message(message, tag)
+        except:
+            pass
+        self.log_text.repaint()  # Ensure GUI updates
+        self.after(100, self.process_log_queue)  # Check queue periodically
+
+    def after(self, ms, func):
+        from PySide6.QtCore import QTimer
```
kjrush: Nit: Might be best to move this import to the top with the rest of the imports.

GUI/src/gui_qt.py
```diff
@@ -0,0 +1,458 @@
...
+        self.browse_button.clicked.connect(self.browse_folder)
+        self.toolbar.addWidget(self.browse_button)
+        
+        # Open CARMA Logs button
+        self.carma_logs_button = QPushButton("Open CARMA Logs")
+        self.carma_logs_button.clicked.connect(lambda: self.open_logs("/opt/carma/logs"))
+        self.toolbar.addWidget(self.carma_logs_button)
+        
+        # Open CARMA-Simulation Logs button
+        self.carma_sim_logs_button = QPushButton("Open CARMA-Simulation Logs")
```
kjrush: This is the CDASim mosaic logs folder right? I would name it as such. Maybe "CDASim Logs" CARMA-Simulation is a name we've moved away from so it would be good to avoid it to minimize confusion.

GUI/src/gui_qt.py
```diff
@@ -0,0 +1,458 @@
...
+        
+        self.setLayout(layout)
+        
+    def get_selected(self):
+        return self.combo.currentText() if self.result() == QDialog.Accepted else None
+
+class SimulatorGUI(QMainWindow):
+    def __init__(self):
+        super().__init__()
+        self.setWindowTitle("CDA Simulator GUI")
```
kjrush: Would recommend "CDASim GUI" for branding consistency.

GUI/src/gui_qt.py
```diff
@@ -0,0 +1,458 @@
...
+        self.setMinimumSize(600, 400)
+        
+        self.repo_folder = None
+        self.selected_folder = None
+        self.config_folder = None
+        self.start_script = None
+        self.stop_script = None
+        self.docker_compose_file = None
+        self.build_script = None
+        self.log_queue = Queue()  # For multiprocessing log messages
```
kjrush, line 64: Is there a reason to use a multiprocessing Queue here over just a standard python list with `append()` and `pop()` being the queue functions? Not sure it makes any difference, but curious if there's a reason to prefer the Queue class over a list-primitive.

GUI/src/gui_qt.py
```diff
@@ -0,0 +1,458 @@
...
+        from PySide6.QtCore import QTimer
+        QTimer.singleShot(ms, func)
+
+    def open_logs(self, path):
+        try:
+            if not os.path.exists(path):
+                self.log_message(f"Error: Log directory {path} does not exist.", "error")
+                return
+            self.log_message(f"Opening log directory: {path}")
+            subprocess.run(["xdg-open", path], check=True)
```
kjrush, line 167: Does this block the main UI thread while the file browser is open? Not a huge deal, especially for this initial MVP build, but could be inconvenient in the final implementation if so.

GUI/src/gui_qt.py
```diff
@@ -0,0 +1,458 @@
...
+
+    def on_config_selected(self, config):
+        if not config:
+            self.log_message("No config selected.")
+            return
+        
+        self.selected_folder = os.path.join(self.repo_folder, config)
+        self.log_message(f"Selected config: {config}")
+        
+        self.docker_compose_file = os.path.join(self.selected_folder, "docker-compose.yml")
```
kjrush, line 225: Again, not necessary to change for this MVP version but would be nice to validate that these files exist at this level (or before showing them in the selectable list even), rather than when the user attempts to actually use the config.

GUI/src/gui_qt.py
```diff
@@ -0,0 +1,458 @@
...
+            return False
+        
+        try:
+            # Copy map
+            dest_maps_dir = "/opt/carma/maps"
+            os.makedirs(dest_maps_dir, exist_ok=True)
+            dest_map = os.path.join(dest_maps_dir, "vector_map.osm")
+            if os.path.exists(dest_map):
+                now = datetime.now().strftime("%m%d%Y%H%M%S")
+                backup = os.path.join(dest_maps_dir, f"vector_map_{now}.osm.backup")
```
kjrush: Not necessary for current MVP version, but would be good to actually compare here not just if `dest_map` exists, but if `dest_map` and the desired map are actually different. This could possibly end up creating a bunch of `vector_map_*.osm.backup` files that are identical copies of each other if this is called multiple times.

kjrush (dismissed):
Just minor comments for future stuff, feel free to merge!

GUI/src/gui_qt.py
chengyuan0124: The code uses `multiprocessing.Process` to run the `start_simulation` and `stop_simulation` scripts in separate processes to avoid blocking the main thread. In a multiprocessing setup, each process has its own memory space, and standard Python lists, from what I understanding, are not shared between processes.

GUI/src/gui_qt.py
chengyuan0124: I think the file validation is handled in the `browse_folder` function at line 172, if we're referring to the same functionality. Based on your previous suggestion, when a user selects a `cdasim-config` folder, the system should automatically populate the dropdown list GUI with the validated `cdasim-config` contents. The validated `cdasim-config` must include the following:

1. docker-compose.yml
2. build-image.sh
3. cdasim_config folder
4. Start/stop simulation scripts
5. MAP folder containing .osm files
6. route_config folder containing route .csv files

GUI/src/gui_qt.py
chengyuan0124: Added a function to compare the .osm under `cdasim_config/MAP` with all maps under `/opt/carma/maps`, if matched then will reuse it

GUI/src/gui_qt.py
chengyuan0124: It does block but only from the moment the `Open CARMA Logs` or `Open CDASim Logs` button is clicked until the file browser window pops up.

kjrush (approved):

PR #252, opened 2025-09-11
Cdas 69 mosaic xmlrpc client implementation

Lazer2077:
# PR Details
## Description
This is the implementation for the Carla XMLRPC Client 

## Related Jira Key
CDAS-69 Ambassadors for MOSAIC Integration

## Motivation and Context
The CARLA-MOSAIC integration currently uses a TraCI-based bridge for communication between CARLA and MOSAIC, managing actors (vehicles, pedestrians) and traffic lights. We have recently implemented an XML-RPC Server in CARLA to enable a more efficient, dedicated connection. To fully transition from the TraCI-based bridge, an XML-RPC Client must be implemented in the CARLA Ambassador within MOSAIC. This client will interact with the XML-RPC Server to handle bidirectional commands for actor lifecycle (create, update, destroy) and traffic light state updates, ensuring seamless synchronization between SUMO and CARLA simulations without reliance on TraCI.

## How Has This Been Tested?
Pass the docker build test

## Types of changes
Mainly change dir: 
1. cdasim/co-simulation/fed/mosaic-carla
2. cdasim/co-simulation/lib/mosaic-interactions

List of files been modified:
1. `CarlaAmbassador.java`: processing the interaction from rti
2. `pom.xml`: Add the build package
3. `CarlaXmlRpcClient.java`: register the corresponding Server function 

Related Interaction will be implemented and extend on the `rti` intereaction, list of file been added:

1. `CarlaActorRequest.java`
5. `CarlaActorResponse.java`
6. `CarlaTrafficLightRequest.java`
7. `CarlaTrafficLightResponse.java`

## Checklist:

- [x] I have added any new packages to the sonar-scanner.properties file


co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/carlaconnect/CarlaMultiXmlRpcManager.java
chengyuan0124, line 1: Do you think if it a good idea to remove all CARLA related functions from CarlaMutiXmlRpcManager.java, such as `createSensor`, `advanceSimulation`, `getAllActors` since they've been developed in CarlaXmlRpcClient.java already. In CarlaAmbassador.java, we can call `CarlaMultiXmlRpcManager.getClient(serverType).createSensor(...)`, for example, directly to use the function from CarlaXmlRpcManager or is there any obstacle for this approach?

Lazer2077: Ok, will do

co-simulation/fed/mosaic-sumo/src/main/java/org/eclipse/mosaic/fed/sumo/ambassador/SumoAmbassador.java
chengyuan0124, line 1: It appears you are considering adding a VehicleUpdates handler in the SUMO Ambassador. The SUMO Ambassador extends AbstractSumoAmbassador, which already implements a VehicleUpdates interaction handler [here](https://github.com/usdot-fhwa-stol/cdasim/blob/818a84eaf59edb2435b4a15100a0abab9b69dc30/co-simulation/fed/mosaic-sumo/src/main/java/org/eclipse/mosaic/fed/sumo/ambassador/AbstractSumoAmbassador.java#L557). This handler does not need to be modified or duplicated in the SUMO Ambassador, as it already fully supports vehicle synchronization for external vehicles. In [receiveInteraction(VehicleUpdates)](https://github.com/usdot-fhwa-stol/cdasim/blob/818a84eaf59edb2435b4a15100a0abab9b69dc30/co-simulation/fed/mosaic-sumo/src/main/java/org/eclipse/mosaic/fed/sumo/ambassador/AbstractSumoAmbassador.java#L557), the handler stores external vehicle state updates (e.g., position, speed, heading) in the externalVehicleMap [here](https://github.com/usdot-fhwa-stol/cdasim/blob/818a84eaf59edb2435b4a15100a0abab9b69dc30/co-simulation/fed/mosaic-sumo/src/main/java/org/eclipse/mosaic/fed/sumo/ambassador/AbstractSumoAmbassador.java#L566) if it is not existing in the externalVehicleMap and removes vehicles via TraCI if needed. These stored updates are then used in the processTimeAdvanceGrant method, specifically in [setExternalVehiclesToLatestPositions](https://github.com/usdot-fhwa-stol/cdasim/blob/818a84eaf59edb2435b4a15100a0abab9b69dc30/co-simulation/fed/mosaic-sumo/src/main/java/org/eclipse/mosaic/fed/sumo/ambassador/AbstractSumoAmbassador.java#L1234), where the stored position and heading data are applied to SUMO using TraCI’s moveToXY command during each simulation time step. This ensures external vehicle states are synchronized with SUMO. No changes are needed to either the SumoAmbassador or AbstractSumoAmbassador. Instead, ensure the CARLA Ambassador correctly triggers VehicleUpdates interactions with accurate vehicle status data, which the SUMO Ambassador will process automatically.

Lazer2077: OK, I delete the duplicated VehilceUpdates intereaction on the `sumoAmbassador.java`

chengyuan0124: I think it is the same for the traffic light. AbstractSumoAmbassador is using [TrafficLightStateChange](https://github.com/usdot-fhwa-stol/cdasim/blob/084b830141b8c126b45a3bde6c31d4e5c6e97361/co-simulation/fed/mosaic-sumo/src/main/java/org/eclipse/mosaic/fed/sumo/ambassador/AbstractSumoAmbassador.java#L773)

Lazer2077: OK, done

co-simulation/bridge/net_offset_reader.py
chengyuan0124, line 1: I think we move the offset reader to CARLA Ambassador, do we still need this?

co-simulation/bridge/carla_xmlrpc_server.py
chengyuan0124, line 1: what is this xml-rpc server, what is the differences between and 0.10 and 0.9

.vscode/settings.json
chengyuan0124, line 1: Please don't push your vscode setting to git

co-simulation/bridge/carla_v0.9_xmlrpc_server.py
```diff
@@ -64,6 +68,21 @@ def __init__(self, host: str = 'localhost', port: int = 8090,
...
+        self.default_extent_x = 2.0
+
+        # External-to-CARLA coordinate transform settings (SUMO/MOSAIC frame → CARLA frame)
+        # - input_frame: 'sumo' applies BridgeHelper-like conversion (y inversion, yaw - 90 deg, offset)
+        # - net_offset_xy: offset from SUMO net (x, y) applied before handedness flip
+        self.input_frame: str = 'sumo'
+        
+        # Try to read net offset from Town04.net.xml, fallback to hardcoded value
+      
+        self.net_offset_xy: Tuple[float, float] = (503.02, 423.76)
```
chengyuan0124: I am not sure we want to have offset here, for example, if we want to integrate other simulation in the future, this default will cause the problem since this is specific for SUMO

co-simulation/bridge/carla_v0.9_xmlrpc_server.py
```diff
@@ -64,6 +68,21 @@ def __init__(self, host: str = 'localhost', port: int = 8090,
...
+        # Flag to track if first actor has been spawned (for automatic spectator switching)
+        self.first_actor_spawned = False
+
+        # Default extent_x for vehicle center calculation
+        self.default_extent_x = 2.0
+
+        # External-to-CARLA coordinate transform settings (SUMO/MOSAIC frame → CARLA frame)
+        # - input_frame: 'sumo' applies BridgeHelper-like conversion (y inversion, yaw - 90 deg, offset)
+        # - net_offset_xy: offset from SUMO net (x, y) applied before handedness flip
+        self.input_frame: str = 'sumo'
```
chengyuan0124: Does it mean it only supports 1 simulation in a run? Why CARLA XML-RPC server need to know about SUMO information?

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
chengyuan0124, line 1: Why we add manual_control.py here? manual_control.py is the script to check it is working, we don't really need to include it here

co-simulation/bridge/carla_xmlrpc_server.py
Lazer2077: I aim to integrate this together, currently I will delete and archive

.vscode/settings.json
Lazer2077: OK, sorry about that

co-simulation/bridge/carla_v0.9_xmlrpc_server.py
Lazer2077: `input_frame` can be changed through the api, and it just a flag determine what should server handling the vehicle transformation

co-simulation/bridge/net_offset_reader.py
Lazer2077: OK, I will implemented offset reader on CARLA Ambassador

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
Lazer2077: You mean I added manual_control on the `cala_v0.10_xmlrpc_server.py`?

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/carlaconnect/CarlaXmlRpcClient.java
```diff
@@ -30,79 +31,1344 @@
...
+        return changes;
+    }
+
+    /**
+     * Check if actor state has changed between two states
+     * @param previousState Previous actor state
+     * @param currentState Current actor state
+     * @return true if state has changed
+     */
+    private boolean hasActorStateChanged(Map<String, Object> previousState, Map<String, Object> currentState) {
```
kjrush: Is there a reason for this being a Map<String, Object> instead of a proper class defition for `ActorState` or something similar? If the key-types here are truly heterogeneous then there's almost certainly a better way to represent this data other than as raw Object types. I see this pattern (use of raw Object) in a few places in this file, I'm assuming that's because we're getting untyped data from the XML-RPC interface (maybe I'll figure out more later in my review, I'll come back to this comment if so). Ideally in that case we should be converting those to proper types and classes at the earliest possible moment, throwing exceptions as appropriate if something happens incorrectly. Passing Object types around just subverts much of the type system's value and makes it more likely for us to introduce errors or maintenance difficulties.

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
```diff
@@ -144,6 +190,92 @@ def _sim_timestamp(self) -> float:
...
+            # Transform to CARLA left-handed system (invert Y)
+            return carla.Location(x_off, -y_off, z_off)
+        # Assume already in CARLA world coordinates
+        return carla.Location(x_in, y_in, z_in)
+
+    def _to_carla_rotation(self, rotation_seq: List[float]) -> carla.Rotation:
+        try:
+            pitch_in = float(rotation_seq[0]) if len(rotation_seq) > 0 else 0.0
+            yaw_in = float(rotation_seq[1]) if len(rotation_seq) > 1 else 0.0
+            roll_in = float(rotation_seq[2]) if len(rotation_seq) > 2 else 0.0
```
kjrush: Nitpick: Not generally a fan of using in-band values to represent out-of-band statuses. I.e. 0.0 is a valid pitch/yaw/roll, but based on the way this function is setup I can't distinguish "yaw not present" from "yaw is 0.0".

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
```diff
@@ -144,6 +190,92 @@ def _sim_timestamp(self) -> float:
...
+            z_off = z_center
+            
+            # Transform to CARLA left-handed system (invert Y)
+            return carla.Location(x_off, -y_off, z_off)
+        # Assume already in CARLA world coordinates
+        return carla.Location(x_in, y_in, z_in)
+
+    def _to_carla_rotation(self, rotation_seq: List[float]) -> carla.Rotation:
+        try:
+            pitch_in = float(rotation_seq[0]) if len(rotation_seq) > 0 else 0.0
```
kjrush: This type conversion doesn't make a ton of sense to me, we're converting to float forcefully, but the type hint above already states the expected input is a List of floats, so `rotation_seq[i]` should be a float already no?

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
```diff
@@ -144,6 +190,92 @@ def _sim_timestamp(self) -> float:
...
+                    yaw_in = float(rotation_seq[1]) if len(rotation_seq) > 1 else 0.0
+                    pitch_in = float(rotation_seq[0]) if len(rotation_seq) > 0 else 0.0
+                    
+                    # From front-center-bumper to center (SUMO reference system)
+                    # Reference: http://sumo.sourceforge.net/userdoc/Purgatory/Vehicle_Values.html#angle
+                    yaw = -1 * yaw_in + 90
+                    x_center = x_in - math.cos(math.radians(yaw)) * extent_x
+                    y_center = y_in - math.sin(math.radians(yaw)) * extent_x
+                    z_center = z_in - math.sin(math.radians(pitch_in)) * extent_x
+                except Exception:
```
kjrush: What exceptions are we anticipating here? We're pretty aggressively checking for None and bounds checking the list accesses, I don't think these math operations can fail, is there some other concern?

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
```diff
@@ -83,6 +94,34 @@ def __init__(self, host: str = 'localhost', port: int = 8090,
...
+                                       base_transform.location.y + dy,
+                                       base_transform.location.z + dz),
+                        base_transform.rotation
+                    )
+                    actor = self.world.try_spawn_actor(bp, t)
+                    if actor is not None:
+                        return actor
+                except Exception:
+                    continue
+        return None
```
kjrush, line 134: Unclear if I haven't reached where this is used, but a failure to initialize an actor despite multiple attempts should be logged or at least printed to stdout/stderr.

.cursor/debug.log
kjrush, line 1: Lets make sure to clean up any local workspace/IDE files before committing.

co-simulation/bridge/__pycache__/manual_control.cpython-311.pyc
kjrush, line 1: Why do we have a compiled binary file committed into the repo? Is this associated with a source file that needs to be reviewed?

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
```diff
@@ -75,6 +77,15 @@ def __init__(self, host: str = 'localhost', port: int = 8090,
 
         self.lock = threading.RLock()
 
+        # External-to-CARLA coordinate transform settings (SUMO/MOSAIC frame → CARLA frame)
+        # - input_frame: 'sumo' applies BridgeHelper-like conversion (y inversion, yaw - 90 deg, offset)
+        # - net_offset_xy: offset from SUMO net (x, y) applied before handedness flip
+        self.input_frame: str = 'sumo'
```
kjrush: Nitpick, not sure I love string as a type for this, but probably not worth changing. If there's only two possible values, a bool would make more sense `is_sumo_coordinate_frame` or something. Otherwise an enum type with entries for each possible coordinate system, or just directly taking in the specific coordinate transform parameters might be cleaner.

Strings are just easy to introduce errors with since there's very little assurance of correctness that the language itself can provide.

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
```diff
@@ -127,11 +166,18 @@ def _register_methods(self):
...
+        # Spectator / camera utilities
+        self.server.register_function(self.set_spectator_to_actor, 'set_spectator_to_actor')
+
         # Maps
         self.server.register_function(self.get_map_name, 'get_map_name')
         self.server.register_function(self.get_available_maps, 'get_available_maps')
         self.server.register_function(self.load_map, 'load_map')
 
+        # Coordinate transform configuration
+        self.server.register_function(self.set_input_frame_mode, 'set_input_frame_mode')
```
kjrush: How is this used in the ambassador side? Is there ever a case where the input frame isn't sumo?

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
```diff
@@ -144,6 +190,92 @@ def _sim_timestamp(self) -> float:
         except Exception:
             return 0.0
 
+    # ----- Coordinate transforms (external → CARLA) -----
+    def _to_carla_location(self, location_seq: List[float], rotation_seq: List[float] = None, extent_x: float = 0.0) -> carla.Location:
+        try:
+            x_in = float(location_seq[0])
+            y_in = float(location_seq[1])
+            z_in = float(location_seq[2]) if len(location_seq) > 2 else 0.0
+        except Exception:
```
kjrush: What exception are we anticipating here?

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
```diff
@@ -144,6 +190,92 @@ def _sim_timestamp(self) -> float:
...
+            roll_in = float(rotation_seq[2]) if len(rotation_seq) > 2 else 0.0
+        except Exception:
+            pitch_in, yaw_in, roll_in = 0.0, 0.0, 0.0
+
+        if self.input_frame == 'sumo':
+            # SUMO → CARLA yaw mapping per BridgeHelper: yaw_carla = yaw_sumo - 90
+            return carla.Rotation(pitch_in, yaw_in - 90.0, roll_in)
+        return carla.Rotation(pitch_in, yaw_in, roll_in)
+
+    def _to_carla_velocity(self, velocity_seq_or_dict: Any) -> carla.Vector3D:
```
kjrush: If it works, it works, but this function feels very strange from a type-system perspective. It implies we'd have a case where we're not sure whether or not the value we have is a dict or seq, but we know it contains velocity data. Which seems like a code smell to me that we may not be managing out types correctly

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
```diff
@@ -144,6 +190,92 @@ def _sim_timestamp(self) -> float:
...
+
+        if self.input_frame == 'sumo':
+            # Flip Y to match CARLA left-handed axes
+            return carla.Vector3D(vx, -vy, vz)
+        return carla.Vector3D(vx, vy, vz)
+
+    # ----- Coordinate transform configuration -----
+    def set_input_frame_mode(self, mode: str) -> bool:
+        try:
+            mode_l = str(mode).strip().lower()
```
kjrush: As before, we've already type hinted this `mode` value as a str, why cast to str explicitly like this?

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
```diff
@@ -144,6 +190,92 @@ def _sim_timestamp(self) -> float:
...
+            if mode_l in ('sumo', 'carla'):
+                self.input_frame = 'sumo' if mode_l == 'sumo' else 'carla'
```
kjrush: This ternary operator is redundant. Line 265 establishes that `mode_l` is either "sumo" or "carla" already. `self.input_frame = mode_l` achieves the same thing as the current ternary operator but without branching. Since if `mode_l` is "sumo", then it sets input frame to sumo, and if mode_l is carla it sets it to carla, and we've already ruled out that it could be anything else.

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
```diff
@@ -194,13 +326,36 @@ def connect(self) -> bool:
                     self.client = carla.Client(self.carla_host, self.carla_port)
                     self.client.set_timeout(10.0)
                 self.world = self.client.get_world()
+                
+                # Set CARLA simulation to passive mode (synchronous mode)
+                settings = self.world.get_settings()
+                settings.synchronous_mode = True
+                settings.fixed_delta_seconds = self.phase
```
kjrush: Seeing how this value is used, I'm not sure phase is a good name for it. `delta_t` or `timestep_size` might be better or something.

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
```diff
@@ -194,13 +326,36 @@ def connect(self) -> bool:
...
                 try:
                     current_map = self.world.get_map().name
                 except Exception:
                     current_map = "<unknown>"
-                logger.info("Connected to CARLA %s at %s:%s | map=%s",
+                logger.info("Connected to CARLA %s at %s:%s | current map=%s",
                             CARLA_VERSION, self.carla_host, self.carla_port, current_map)
                 
+                # Automatically load Town04 if not already loaded
+                if current_map != "Town04":
```
kjrush: I don't think we want to hard code loading Town04, we might run different scenarios on various CARLA maps.

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
```diff
@@ -418,14 +648,68 @@ def update_actor_velocity(self, actor_key: ActorKey, velocity: List[float]) -> b
     def get_all_actors(self) -> Dict[str, Dict[str, Any]]:
         try:
             with self.lock:
```
kjrush: Maybe it'll be clear as I continue reading, but are we multithreading here? I haven't yet seen it. Is the XML RPC server itself handling the threading on a per-RPC basis?

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
```diff
@@ -327,16 +482,66 @@ def spawn_actor(self, actor_type: str, actor_id: str,
...
                 self.actor_blueprints[actor_id] = bp
-                return True
+                
+                # Only switch spectator to the first spawned actor
+                if len(self.actors) == 1:  # Only for the first vehicle
+                    try:
+                        # Switch spectator to follow the first spawned actor with proper offset
+                        # Use the actual actor object for more reliable positioning
+                        self._set_spectator_to_actor_object(actor, 'follow', 8.0, 3.0, -20.0)
+                        print(f"========Spectator switched to follow first actor: {actor_id}========")
```
kjrush: Would recommend converting all print statements to log statements using the logger object.

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
kjrush, line 1: Most of my comments here can be interpreted as nitpicks if this has been thoroughly tested. The combination of type hints and lots of forced typecasting stands out to me as kind of odd and likely not the most maintainable way to implement this. Additionally, the prevalance of the

```
try:
expect Exception:
  continue
```

or similar pattern is a bit concerning, especially given that it surrounds a fair number of blocks that don't seem like they should throw exceptions. If they're throwing exceptions in practice, and this is a workaround, then we should probably address the underlying issue (likely caused by lack of None-checks or type checks somewhere further upstream in the stack).

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
EricChen-Lei: It would be better to log down exception for debugging purposes

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
EricChen-Lei: It would be better to log down exception for debugging purposes

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
```diff
@@ -144,6 +190,92 @@ def _sim_timestamp(self) -> float:
...
+            return carla.Location(x_off, -y_off, z_off)
+        # Assume already in CARLA world coordinates
+        return carla.Location(x_in, y_in, z_in)
+
+    def _to_carla_rotation(self, rotation_seq: List[float]) -> carla.Rotation:
+        try:
+            pitch_in = float(rotation_seq[0]) if len(rotation_seq) > 0 else 0.0
+            yaw_in = float(rotation_seq[1]) if len(rotation_seq) > 1 else 0.0
+            roll_in = float(rotation_seq[2]) if len(rotation_seq) > 2 else 0.0
+        except Exception:
```
EricChen-Lei: It would be better to log down exception for debugging purposes

co-simulation/bridge/carla_v0.9_xmlrpc_server.py
```diff
@@ -116,11 +174,16 @@ def _register_methods(self):
         self.server.register_function(self.get_sensor_data, 'get_sensor_data')
```
chengyuan0124: For the actor xml-rpc server, we shouldn't include anything about sensor or is there any reason we do this?

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
chengyuan0124, line 1: I know during the testing, it is more focusing on v0.9, but do we do the parallel upgrade to v0.10 when some logic change in v0.9?

co-simulation/bridge/__pycache__/manual_control.cpython-311.pyc
chengyuan0124, line 1: Don't push the python compile file

co-simulation/bridge/net_offset_reader.py
chengyuan0124: I am not sure if we still need this or not. your reply mentioned it is moved to CARLA Ambassador but still having script here

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
```diff
@@ -112,6 +128,37 @@ public class CarlaAmbassador extends AbstractFederateAmbassador {
...
+    public Map<String, String> getSumoToCarlaIdMapping() {
+        return new HashMap<>(sumoToCarlaIdMapping);
+    }
+
+
+    /**
+     * SUMO net offset parsed from scenario .net.xml (x, y) in meters.
+     * Default to Town04 values if parsing fails.
+     */
+    private double[] sumoNetOffsetXY = new double[]{503.02, 423.76};
```
chengyuan0124: Do we need this default offset setting? Will this be overwritten by the value read from file?

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
```diff
@@ -211,101 +290,76 @@ public void initialize(long startTime, long endTime) throws InternalFederateExce
...
-                connectionProcess = Runtime.getRuntime().exec(command, null, new File(path));
-            } catch (Exception ex) {
-                ex.printStackTrace();
-                if (connectionAttempts == 0) {
-                    log.info("Maximum connection attempts reached and connecting to CARLA simulator failed.");
-                } else {
-                    log.warn("Error while connecting to CARLA simulator. Retrying.");
+                // Add sensor library server
+                // convert string null to null
+                if ("null".equalsIgnoreCase(carlaConfig.carlaSensorLibRPCUrl)) {
```
chengyuan0124, line 316: Not sure what is this line and line 307 do, if it is necessary to convert that from string to null, then we will have problem in line 297 since that is raw data without converting

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
```diff
@@ -211,101 +290,76 @@ public void initialize(long startTime, long endTime) throws InternalFederateExce
...
+                
+                if (multiXmlRpcManager.getClient(CarlaXmlRpcClient.ServerType.SENSOR_LIB) == null &&
+                    multiXmlRpcManager.getClient(CarlaXmlRpcClient.ServerType.ACTOR_LIB) == null) {
+                    throw new InternalFederateException("No XML-RPC servers configured for multi-server mode");
+                }
+                
+            } catch (MalformedURLException m) {
+                throw new InternalFederateException("Carla Ambassador initialization failed due to invalid XML-RPC server URLs! Check carla_config.json!");
             }
+        } else {
```
chengyuan0124: This `else` is pointless because these conditions are never true due to the logic of the parent `if`

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
```diff
@@ -349,32 +402,232 @@ public synchronized void processTimeAdvanceGrant(long time) throws InternalFeder
...
         try {
-            if ( time == 0 ) {
-                // Try to connect to CARLA CDA Sim Adapter on first timestep
-                carlaXmlRpcClient.connect(60);
+            if (time == 0) {
+                // Try to connect to XML-RPC servers on first timestep
+                if (multiXmlRpcManager != null) {
+                    multiXmlRpcManager.connectAll(60);
+                } else if (carlaXmlRpcClient != null) {
+                    carlaXmlRpcClient.connect(60);
```
chengyuan0124: What is carlaXmlRpcClient? is that for sensor xml rpc server? but I thought it will be saved into `multixmlrpcmanager`?

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
```diff
@@ -349,32 +402,232 @@ public synchronized void processTimeAdvanceGrant(long time) throws InternalFeder
...
-                        // Convert nanosecond timestamp to millisecond timestamp
-                        interaction.getDetectedObject().setTimestamp((int)(time/1e6));
-                        detectedObjectInteractions.add(interaction);
+            // Advance CARLA simulation by one tick before polling sensors/actors
+            try {
+                CarlaXmlRpcClient tickClient = null;
+                if (multiXmlRpcManager != null) {
+                    tickClient = multiXmlRpcManager.getClient(CarlaXmlRpcClient.ServerType.ACTOR_LIB);
+                } else {
+                    tickClient = carlaXmlRpcClient;
```
chengyuan0124: same question, what is carlaXmlRpcClient? if it is sensor xml rpc, I don't recall it is calling tick

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
```diff
@@ -462,45 +720,405 @@ public boolean isTimeRegulating() {
...
-     * Trigger a new CarlaTraciRequest, SimulationStep or ExternalMessage
-     * interaction
-     *
-     * @param length  command length
-     * @param command command
+     * Internal method to trigger simulation step coordination.
+     * This replaces the external SimulationStep interaction with internal logic.
      */
-    public synchronized void triggerInteraction(int length, byte[] command) throws InternalFederateException {
+    private void triggerInternalSimulationStep() {
```
chengyuan0124: We can remove this function, this is not used anymore

chengyuan0124 (changes requested):
Released my comment for xml-rpc server and CARLA Ambassador, will continue to review the reset

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
Lazer2077: OK, I will log down these stuff on carla_v0.10_xmlrpc_server.py

co-simulation/bridge/__pycache__/manual_control.cpython-311.pyc
Lazer2077: I will delete this stuff

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
Lazer2077: done

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
Lazer2077: delete it

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
Lazer2077: I'm not sure it will be overwritten or not; I can conduct test about that. I remember you implemented the reading stuff

co-simulation/bridge/carla_v0.9_xmlrpc_server.py
Lazer2077: I will delete them

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
chengyuan0124: No I didn't implement it, what I did was just adding parameter in carla_conifg.json in 2 of the pushes [[1]](https://github.com/usdot-fhwa-stol/cdasim/pull/252/commits/fc2f4b82d0691ffc5fcbfa1ba17553e399e9255f) [[2]](https://github.com/usdot-fhwa-stol/cdasim/pull/252/commits/1513889831e9c701dde7d5c79786092efaab83d3)

co-simulation/fed/mosaic-carla/src/test/resources/carla_config.json
chengyuan0124, line 1: Not sure what is this file, is this necessary?

co-simulation/fed/mosaic-carla/src/test/resources/carla_config_debug.json
chengyuan0124, line 1: Not sure what is this file, is this necessary?

co-simulation/fed/mosaic-carla/src/test/resources/carla_config_with_map.json
chengyuan0124, line 1: Not sure what is this file, is this necessary?

co-simulation/fed/mosaic-sumo/src/main/java/org/eclipse/mosaic/fed/sumo/ambassador/AbstractSumoAmbassador.java
```diff
@@ -559,6 +560,43 @@ private synchronized void receiveInteraction(VehicleUpdates vehicleUpdates) thro
             return;
         }
 
+        // Handle added vehicles from external simulators (like CARLA)
```
chengyuan0124: Do not infer external vehicles from `VehicleUpdates`.
Instead, CARLA Ambassador should explicitly trigger a `VehicleFederateAssignment` firstly when a external vehicle created (e.x, in CARLA) before sending `VehicleUpdates` for that vehicle.”

chengyuan0124 (changes requested):
Completed my initial review. Some logic in the CARLA–SUMO interaction needs to be adjusted. Specifically, when CARLA spawns a vehicle, it should trigger a `VehicleFederateAssignment` first, before sending any `VehicleUpdates`

co-simulation/bridge/carla_v0.9_xmlrpc_server.py
```diff
@@ -64,6 +70,12 @@ def __init__(self, host: str = 'localhost', port: int = 8090,
 
         self.lock = threading.RLock()
```
chengyuan0124: Why we are using lock in xmlrpc_server.py? It seems to me there is no multi-thread method in this python script

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
Lazer2077: It has been confirmed that the v0.10 has been updated

co-simulation/bridge/carla_v0.9_xmlrpc_server.py
Lazer2077: what about the sensor lib? do we need lock to prevent conflict?

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
Lazer2077: It just ensure that every format of Url can be processed

co-simulation/bridge/carla_v0.9_xmlrpc_server.py
chengyuan0124: The carla_v0.9_xmlrpc_server.py is only for actor, it doesn't need to include the sensor support, sensor support function is in the carla-sensor-lib repo, two of them are different xml-rpc server. Is there any reason why you are considering the sensor in the actor-xml-rpc?

co-simulation/fed/mosaic-sumo/src/main/java/org/eclipse/mosaic/fed/sumo/ambassador/AbstractSumoAmbassador.java
chengyuan0124: did you update CARLA Ambassador to send `VehicleFederateAssignment` ? I only see you delete this section in `AbstractSumoAmbassador`

kjrush (commented):
Leaving partial comments, wont have time to finish this review within the sprint. I'll defer to @chengyuan0124 on whether/how these get resolve and if similar issues exist elsewhere, I've only had time to review this one file mostly.

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
Lazer2077: It is just temp solution for the Town04 test, I will implement this function later.

co-simulation/bridge/carla_v0.9_xmlrpc_server.py
Lazer2077: OK, i delete the lock implementation

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
Lazer2077: replaced

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
Lazer2077: It is confirmed that it is working after setting the offset to 0,0

co-simulation/fed/mosaic-carla/src/test/resources/carla_config_with_map.json
Lazer2077: deleted

co-simulation/fed/mosaic-sumo/src/main/java/org/eclipse/mosaic/fed/sumo/ambassador/AbstractSumoAmbassador.java
Lazer2077: updated

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
Lazer2077: removed

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
Lazer2077: replaced

.cursor/debug.log
Lazer2077: done

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
Lazer2077: deleted

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
Lazer2077: deleted

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
Lazer2077: deleted

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
Lazer2077: deleted

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
chengyuan0124: What I meant was that this logic is semantically backwards. Shouldn’t the string-to-null conversion happen before this `if (carlaConfig.carlaSensorLibRPCUrl != null || carlaConfig.carlaActorLibRPCUrl != null)` check? You’re trying to ensure the format is correct by converting `"null"` (as a string) into an actual `null`, but `carlaConfig.carlaSensorLibRPCUrl` and `carlaConfig.carlaActorLibRPCUrl` are being validated only after this condition runs. This makes the control flow confusing, because the if condition depends on values that haven’t been sanitized yet.

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
chengyuan0124: @Lazer2077

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
chengyuan0124: @Lazer2077

co-simulation/bridge/carla_v0.10_xmlrpc_server.py
chengyuan0124: @Lazer2077

Lazer2077: _to_carla_rotation this function no longer used

co-simulation/fed/mosaic-carla/src/main/java/org/eclipse/mosaic/fed/carla/ambassador/CarlaAmbassador.java
Lazer2077: I delete it

chengyuan0124 (approved):

PR #254, opened 2025-09-23
CDAS 43 scenario generator for cda sim automated testing tool

chengyuan0124:
# PR Details
## Description

The script will orchestrate the deployment of Docker Compose configurations for sequential simulation runs, eliminating manual setup. The generator will use YAML configurations to produce the bash script using JinJa2, which automates the deployment of CDASim, CARMA Streets, and vehicle services (platform + bridge) by navigating directories, creating networks, and executing docker compose up with predefined docker-compose.yml and .env files.
## Related Jira Key

[CDAS-43](https://usdot-carma.atlassian.net/browse/CDAS-43)
## Types of changes

## Checklist:

[CDAS-43]: https://usdot-carma.atlassian.net/browse/CDAS-43?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scenario_generator/config/parameters/parameter.yaml
```diff
@@ -0,0 +1,40 @@
...
+street_config_dir: <path-to-street-config>
+carma_config_dir: <path-to-carma-config>
+env_settings:
+  cdasim:
+    DOCKER_ORG: usdotfhwastol
+    DOCKER_TAG: carma-system-4.10.0
+    PROJECT_NAME: cdasim
+  vehicles:
+    - PROJECT_NAME: vehicle1
+      CARMA_CONFIG_PATH: /path/to/carma-config
```
kjrush, line 28: Is there a reason for using a directory for the vehicle CARMA configs instead of images?

scenario_generator/config/parameters/parameter.yaml
```diff
@@ -0,0 +1,40 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#  License for the specific language governing permissions and limitations under
+#  the License.
+
+#  NOTE: This is a temporary file, further planning and implementation will be conducted in the following sprint
+
+cdasim_config_dir: <path-to-carma-config>
+street_config_dir: <path-to-street-config>
+carma_config_dir: <path-to-carma-config>
```
kjrush, line 19: what's the difference between these values here, and the values below under the vehicle heading?

scenario_generator/config/parameters/parameter.yaml
```diff
@@ -0,0 +1,40 @@
...
+env_settings:
+  cdasim:
```
kjrush, line 21: Is this changeset supposed to account for things like selecting map/scenario within CDASim?

scenario_generator/config/parameters/parameter.yaml
```diff
@@ -0,0 +1,40 @@
...
+    DOCKER_ORG: usdotfhwastol
+    DOCKER_TAG: carma-system-4.10.0
+    PROJECT_NAME: cdasim
+  vehicles:
+    - PROJECT_NAME: vehicle1
```
kjrush, line 27: Is the vehicle spawn point configurable?

scenario_generator/config/parameters/parameter.yaml
```diff
@@ -0,0 +1,40 @@
...
+      CARMA_CONFIG_PATH: /path/to/carma-config
+      DOCKER_ORG: usdotfhwastol
+      DOCKER_TAG: carma-system-4.10.0
+    - PROJECT_NAME: vehicle2
+      CARMA_CONFIG_PATH: /path/to/carma-config
+      DOCKER_ORG: usdotfhwastol
+      DOCKER_TAG: carma-system-4.10.0
+  streets:
+    - PROJECT_NAME: street1
```
kjrush, line 54: Likewise with the spawn point above, where does the RSU for this instance get created? Which intersection does it get connected to in SUMO?

kjrush (commented):
Not sure I fully understand the scope of this feature set, if my comments here are out of scope just let me know.

scenario_generator/config/templates/sim_launch_template.sh.j2
```diff
@@ -0,0 +1,55 @@
...
+# Create shared sim_net
+docker network create sim_net || true
+
+# Launch CDASim
+cd "$CDASIM_CONFIG_DIR" || { echo "Failed to cd to $CDASIM_CONFIG_DIR"; cd "$CURRENT_DIR"; exit 1; }
+source .env.cdasim
+docker compose -f docker-compose.yml --env-file .env.cdasim -p "${PROJECT_NAME}" up -d
+
+# Launch vehicles (carma_config includes platform)
+for i in $(seq 1 $NUM_VEHICLES); do
```
kjrush: Just a thought, it doesn't look like we need to do it just now, but as you get to implementing more options for vehicles and infrastructure instances it's worth considering whether iterating in bash makes more sense than iterating at the template level. Might simplify some things, like you wouldn't need to pass in NUM_VEHICLES since the template engine already knows that value. https://jinja.palletsprojects.com/en/stable/templates/#for

scenario_generator/config/templates/sim_launch_template.sh.j2
```diff
@@ -0,0 +1,55 @@
...
+  cd "$CARMA_CONFIG_DIR" || { echo "Failed to cd to $CARMA_CONFIG_DIR"; cd "$CURRENT_DIR"; exit 1; }
+  source .env.vehicle_${i}
+  # Create platform network for this vehicle
+  docker network create ${PROJECT_NAME}_platform_net || true
```
kjrush: What's the intent of the `|| true` here?

scenario_generator/config/templates/sim_launch_template.sh.j2
```diff
@@ -0,0 +1,55 @@
...
+docker network create sim_net || true
+
+# Launch CDASim
+cd "$CDASIM_CONFIG_DIR" || { echo "Failed to cd to $CDASIM_CONFIG_DIR"; cd "$CURRENT_DIR"; exit 1; }
+source .env.cdasim
+docker compose -f docker-compose.yml --env-file .env.cdasim -p "${PROJECT_NAME}" up -d
+
+# Launch vehicles (carma_config includes platform)
+for i in $(seq 1 $NUM_VEHICLES); do
+  cd "$CARMA_CONFIG_DIR" || { echo "Failed to cd to $CARMA_CONFIG_DIR"; cd "$CURRENT_DIR"; exit 1; }
```
kjrush: Might be best, if possible, to have a cleanup step here, if we've gotten through some number of projects initialized and fail on the nth one, we should shutdown gracefully rather than leaving the user to manually shutdown the docker compose systems.

scenario_generator/config/templates/sim_launch_template.sh.j2
```diff
@@ -0,0 +1,55 @@
...
+
+# Store current directory
+CURRENT_DIR=$(pwd)
+
+# Create shared sim_net
+docker network create sim_net || true
+
+# Launch CDASim
+cd "$CDASIM_CONFIG_DIR" || { echo "Failed to cd to $CDASIM_CONFIG_DIR"; cd "$CURRENT_DIR"; exit 1; }
+source .env.cdasim
```
kjrush: Is this the proper mechanism for this? If we're sourcing each .env file in order, then we're accumulating environment variables as we go along correct? That might have unintended side effects. I'm not sure the exact mechanism, but we want to load `.env.cdasim` only for cdasim and not have those variables possibly conflict with behaviors in platform/etc. 
Might be a `bash -c ...` type solution to spawn a subshell with a fresh set of environment variables, I can look into it.

kjrush: I think a subshell might the correct approach here. On the line where you want to run the docker-compose command that uses these environment variables do `(source <ENV_FILE_HERE>; docker-compose ...)`, that'll spawn a fresh subprocess and then run the stuff inside the parens. So the parent process environment doesn't get polluted with the newly sourced environment variables from the env file.

scenario_generator/config/templates/sim_launch_template.sh.j2
```diff
@@ -0,0 +1,70 @@
...
+# Launch CARMA Streets
+{% for street in streets %}
+cd "$STREET_CONFIG_DIR" || { echo "Failed to cd to $STREET_CONFIG_DIR"; cleanup; }
+(source .env.street_{{ loop.index }} && docker compose -f docker-compose.yml --env-file .env.street_{{ loop.index }} -p "{{ street.PROJECT_NAME }}" up -d && PROJECT_NAMES+=("{{ street.PROJECT_NAME }}"))
+# Create street network for this instance
+docker network create {{ street.PROJECT_NAME }}_street_net || true
+{% endfor %}
+
+# Return to original directory
+cd "$CURRENT_DIR" || exit 1
```
kjrush, line 71: Would be good to have a final echo here to indicate that the launch has completed successfully

scenario_generator/config/templates/sim_launch_template.sh.j2
```diff
@@ -0,0 +1,70 @@
...
+cd "$CARMA_CONFIG_DIR" || { echo "Failed to cd to $CARMA_CONFIG_DIR"; cleanup; }
+(source .env.vehicle_{{ loop.index }} && docker compose -f docker-compose.yml --env-file .env.vehicle_{{ loop.index }} -p "{{ vehicle.PROJECT_NAME }}" up -d && PROJECT_NAMES+=("{{ vehicle.PROJECT_NAME }}"))
+# Create platform network for this vehicle
+docker network create {{ vehicle.PROJECT_NAME }}_platform_net || true
+{% endfor %}
+
+# Launch CARMA Streets
+{% for street in streets %}
+cd "$STREET_CONFIG_DIR" || { echo "Failed to cd to $STREET_CONFIG_DIR"; cleanup; }
+(source .env.street_{{ loop.index }} && docker compose -f docker-compose.yml --env-file .env.street_{{ loop.index }} -p "{{ street.PROJECT_NAME }}" up -d && PROJECT_NAMES+=("{{ street.PROJECT_NAME }}"))
```
kjrush: Is sourcing the env file and passing it to the --env-file argument redundant? If the argument works as I'd expect it to here, I'd be in favor of just ditching the subshell and source command here.

kjrush (approved):

PR #255, opened 2025-10-14
Cdas 79 carla dind

chengyuan0124:
# PR Details
## Description

The primary goal of this PR is to enable CARLA to launch using the official CARLA Docker image in a Docker-in-Docker (DinD) setup, as CDASim itself runs within a Docker container. This update addresses a version conflict issue. CDASim has been migrated to ROS2 and requires CARLA version 0.10.0. However, CARLA 0.10.0 requires Ubuntu 22.04, while the current NS-3.28 in CDASim is compatible only with Ubuntu 18.04. By leveraging the CARLA Docker image in a DinD configuration, this PR resolves the compatibility issue and ensures seamless integration.

## Related Jira Key

[CDAS-79](https://usdot-carma.atlassian.net/browse/CDAS-79)
## Types of changes

## Checklist:

[CDAS-79]: https://usdot-carma.atlassian.net/browse/CDAS-79?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


co-simulation/patch/CarlaUE4.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# CarlaUE4.sh (or CarlaExecutor.sh, linked as CarlaUE4.sh)
+
+# Retrieve the Docker network name of the current host's container
+NETWORK=$(docker inspect --format '{{range $key, $value := .NetworkSettings.Networks}}{{$key}}{{end}}' $(hostname))
```
kjrush, line 5: Just to confirm my understanding: This should already exist because the docker compose from `./start-simulation` would have already created it?

co-simulation/patch/CarlaUE4.sh
```diff
@@ -0,0 +1,18 @@
...
+  echo "Warning: Could not determine Docker network. Falling back to 'host' network."
+  NETWORK="host"
+fi
+
+# Run the Docker command with the dynamically determined network
+docker run --privileged --gpus all --name carla-server --net="$NETWORK" --env=DISPLAY=$DISPLAY \
+  --env=NVIDIA_VISIBLE_DEVICES=all --env=NVIDIA_DRIVER_CAPABILITIES=all \
+  --env=SDL_VIDEODRIVER=x11 --volume="/tmp/.X11-unix:/tmp/.X11-unix:rw" \
+  --volume="/usr/share/vulkan/icd.d:/usr/share/vulkan/icd.d:ro" \
+  carlasim/carla:0.10.0 bash CarlaUnreal.sh -vulkan -nosound "$@"
```
kjrush, line 18: So, if the CARLA image isn't available locally when the user runs it, we'll download it at this step correct?

If there's a way for us to force that to happen earlier, or at least recommend in documentation that users do it before running the simulation that might be slightly better.

co-simulation/patch/CarlaUE4.sh
chengyuan0124: The CDASim Docker service defined in the docker-compose configuration automatically creates the "simulation" Docker network. Since the CARLA Docker container is not managed by docker-compose, it does not inherently know which network to use. To ensure the CARLA container operates within the same Docker network as CDASim, the network parameter must be retrieved and specified when launching the CARLA container.

co-simulation/patch/CarlaUE4.sh
chengyuan0124: I've added the CARLA docker image pull process in `build-image.sh`

chengyuan0124: This approach may only work when building the Docker image. For users pulling the cdasim image directly from Docker Hub, automating this process might not be feasible. However, we can address this in cdasim-config by including the CARLA Docker image in the Docker Compose file. I'll document this and update it once we work on the cdasim-config repository

kjrush: Yeah, I would probably recommend against doing it in build-image.sh. That's just likely to slow down our CI processes and most of the time not the route that end users are going to use. If it's not practical, that's fine, it's not a requirement. I was just asking to see if it was something that had a straightforward solution.

chengyuan0124: Understood, I've reverted that back

kjrush (approved):

PR #279, opened 2026-08-17
cdas-143-add-reusable-compose-override-support-to-scenario-runner

chengyuan0124:
# PR Details
## Description

This PR improves Scenario Runner topology management, Compose configuration, and data collection.

- Adds centralized network and IP allocation through `topology.json` for CDASim, Platform, Messenger, and Street/V2XHub components.
- Adds layered Compose override support across all components, applying base, user-provided, and Scenario Runner-managed Compose files in order.
- Normalizes legacy Compose service names and references when extracting configuration images.
- Updates generated start/stop scripts to manage networks, configuration containers, selected services, and Compose files.
- Uses each scenario’s label as its collected-data directory name.

These changes keep network and container deployment details out of parameters.yaml and make scenario configurations easier to reuse across multiple component instances.
## Related Jira Key

[CDAS-143](https://usdot-carma.atlassian.net/browse/CDAS-143)

## Types of changes

## Checklist:

[CDAS-143]: https://usdot-carma.atlassian.net/browse/CDAS-143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scenario_runner/config/templates/sim_start_template.sh.j2
```diff
@@ -14,34 +14,30 @@
...
 
-docker compose -f "{{ c.compose_file }}" \
-               --env-file "{{ c.env_file }}" \
+{% for c in scenario %}
+docker compose \
+{% for compose_file in c.compose_files %}               -f "{{ compose_file }}" \
+{% endfor %}               --env-file "{{ c.env_file }}" \
                -p "{{ c.PROJECT_NAME }}" \
-               up -d && PROJECT_NAMES+=("{{ c.PROJECT_NAME }}")
+               up -d{% for service in c.services %} "{{ service }}"{% endfor %}
```
paulbourelly999, line 39: What is the purpose of allowing this script to select services? I think if we want this functionality we should use docker compose profiles instead to limit the amount we couple this tool to the service implementation of deployed tools

scenario_runner/config/templates/sim_stop_template.sh.j2
```diff
@@ -18,21 +18,20 @@ CURRENT_DIR=$(pwd)
...
-
-{% for c in scenario %}
-{% if c.platform_net %}
-docker network rm {{ c.platform_net }} 2>/dev/null || true
-{% endif %}
-{% if c.street_net %}
-docker network rm {{ c.street_net }} 2>/dev/null || true
-{% endif %}
+{% for config_container in config_containers %}
+docker rm -f "{{ config_container.name }}" 2>/dev/null || true
```
paulbourelly999, line 29: Why would we need to remove config images separately from docker compose down. Are these launched separately from the services themselves?

scenario_runner/config/topology.json
paulbourelly999, line 1: While I agree that defining some configuration of virtual networks might be useful I think defining the IPs for each service here is creating far to much coupling on the underlying tools networking.

kjrush: I think I understand your point re: coupling, but I'm not sure I understand what the alternative would be? Is there another layer of abstraction here where it makes more sense to define these network parameters? I think this level of subnet configuration does seem useful (to make sure whatever configuration we want to run that our network solution works for each system) but I agree that it's a lot of detail to go into at this level.

Is there some sort of way for docker networking to describe network relationships between components in a relative sense (e.g., these services are on the same subnet, these services are on adjacent IPs, etc.) and then we can map that into subnets at this level and not worry about the underlying structure per system?

scenario_runner/config/topology.json
```diff
@@ -0,0 +1,192 @@
+{
```
kjrush: Minor nitpick, I would recommend renaming this "network_topology.json" to be more specific about what kind of topology it is.

scenario_runner/scenario_topology.py
```diff
@@ -0,0 +1,373 @@
+#  Copyright (C) 2025 LEIDOS.
```
kjrush, line 1: Again, might recommend renaming to network_topology.py or scenario_network_topology.py something like that for clarity.

scenario_runner/config/topology.json
paulbourelly999: docker is able to dynamically assign IPs and docker networking is able to handle DNS name resolution so that instead of having to keep track of each V2X Hub IP on the SIM network you can just call the container name so v2xhub-1 v2xhub-2 etc.  We use this in several places in V2X Hub docker compose like [here ](https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/f5141471655bda1ee94d3278a89deba4d9cbb283/configuration/docker-compose.yml#L106) and [here ](https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/f5141471655bda1ee94d3278a89deba4d9cbb283/configuration/docker-compose.yml#L16)

scenario_runner/config/compose/carma-cloud.cdasim.yml
```diff
@@ -0,0 +1,26 @@
+# CDASim/Scenario Runner overlay for the repository-local CARMA Cloud Compose.
+# The base repository supplies the application configuration files; this file
+# supplies XIL networking and scenario output paths.
+
+services:
+  carma-cloud:
+    image: "${CARMA_CLOUD_IMAGE}"
+    entrypoint:
+      - /opt/tomcat/bin/catalina.sh
+      - run
```
paulbourelly999: Why does this have 2 entrypoints?

scenario_runner/config/compose/carma-cloud.cdasim.yml
```diff
@@ -0,0 +1,26 @@
...
+    command: []
+    network_mode: !reset null
```
paulbourelly999: NIT: Can we add a comment here that this is intended to override the CARMA Cloud Docker compose network mode: host setting

scenario_runner/config/compose/carma-cloud.cdasim.yml
```diff
@@ -0,0 +1,26 @@
...
+# The base repository supplies the application configuration files; this file
+# supplies XIL networking and scenario output paths.
+
+services:
+  carma-cloud:
+    image: "${CARMA_CLOUD_IMAGE}"
+    entrypoint:
+      - /opt/tomcat/bin/catalina.sh
+      - run
+    command: []
```
paulbourelly999: This seems unnecessary since we do not set command in the carma-cloud repo

scenario_runner/config/parameters/parameters.yaml
```diff
@@ -1,146 +1,117 @@
...
     data_output:
-      output_directory: /opt/carma-simulation/tests/output/highway_merge
+      output_directory: /opt/carma-simulation/tests/output/town10
       collect:
         mosaic_logs: /opt/carma-simulation/logs
         rosbags: /opt/carma/logs
-        v2xhub_logs: /opt/carma/logs/v2xhub
-        carmacloud_logs: /opt/carma/logs/carmacloud
\ No newline at end of file
+        v2xhub_logs: /tmp/cdasim-scenario-runner
```
paulbourelly999: NIT: We may want to at least separate logs by service name so like `/tmp/cdasim-scenario-runner/v2xhub/`

scenario_runner/config/templates/sim_start_template.sh.j2
paulbourelly999: Yeah I suggested we still remove services here and use docker compose profiles when applicable

scenario_runner/config/network_topology_template.json
```diff
@@ -0,0 +1,153 @@
...
+      "network_name_template": "xil_street_{index}_net",
+      "private_endpoints": {
+        "v2xhub": {
+          "env": "V2XHUB_PRIVATE_HOST",
+          "host": "v2xhub"
+        },
+        "zookeeper": {
+          "env": "STREET_INFRASTRUCTURE_HOST",
+          "host": "zookeeper",
+          "shares_network_namespace_with": [
```
paulbourelly999: What is this variable?

scenario_runner/README.md
```diff
@@ -64,14 +67,17 @@ The system automatically:
...
 ## Key YAML Fields
 
-Each test case in `parameters.yaml` defines simulation settings under `env_settings`, with the following key fields for every component (cdasim, vehicle, street):
+Scenario Runner supports ROS 2 components only. The entries under
+`env_settings` use these component fields:
 
 | Field | Description |
 |--------|--------------|
 | **`PROJECT_NAME`** | A unique name for this simulation component (used as the Docker project name). |
+| **`COMPONENT`** | Optional vehicle entry type: `platform` (default) or `messenger`. |
```
paulbourelly999: Lets come up with a generic definition for what a component is.

paulbourelly999 (approved):

